PR #5, opened 2018-10-01
add linknet

n01z3:


models/segmentation/linknet.py
```diff
@@ -0,0 +1,142 @@
+import torch
+import torch.nn as nn
+from torch.autograd import Variable
```
Scitator: to pytorch 0.4.1, please

models/segmentation/unet.py
```diff
@@ -1,5 +1,10 @@
+"""
+Made by @nizhib
+"""
+
 import torch
 from torch import nn
+from torch.autograd import Variable
```
Scitator: to pytorch 0.4.1, please

PR #8, opened 2018-10-03
Tighten up the flake8 rules

cclauss:


cclauss:
Why does __yapf --diff --recursive .__ fail?

cclauss:
Ready for review.  It is good enough for now.

Scitator:
@cclauss speaking about yapf,
looks like this command will do, what you want (ping me, if I am right)
```
- yapf.sh --all
```
I already have this bash script for automatic codestyle check :)

PR #16, opened 2018-11-05
loaders in state&&is_test=True for inference

lightforever:


Scitator:
Hi,

I don't think we need additional hardcoded argument `is_test` during inference. Mainly, you already have it by `**config["data_params"]` params. Here is an example – https://github.com/Scitator/catalyst-examples/blob/master/finetune/data.py#L93.

Could you please provide more info, why do you need all loaders in the RunnerState?

And good catch for inference `callbacks_params`, we definitely need them to be optional.

lightforever:
How do you do?!

I agree about ```is_test``` and I have removed it from the merge request. Thanks for the tip.

I have a huge dataset that can't be processed in many epochs and I have a trick for that in my model.py

```
class ChangeDatasetFileCallback(Callback):
    def on_epoch_end(self, state):
        loader = state.loaders['train']
        loader.dataset.epoch = state.epoch + 1
```

And my Dataset implementation applies this variable to create smaller epochs.
I am sure, It's not the right way, but I didn't find how can we manage huge datasets now(for training purposes).

Scitator:
To address huge datasets problem, it's much better to use Samplers.
You can pick this one (github.com/Scitator/catalyst/blob/master/data/sampler.py#L7) or write you own, that will sample only a part of a dataset per epoch (10% or 25%, for example).
You can also add additional configuration to make it work sequentially, like sample all you dataset in 10 epochs, or random, like sample random 10% of your data each epoch.

lightforever:
This is definitely better.
Thanks for the information.

Only making ```callbacks_params``` optional remains in my pull request.

Scitator:
Thanks!

btw, looking forward for your sampler initialization :)
I think, there are a lot of people with same huge-dataset problem, so you can add one more best practise for them.

PR #17, opened 2018-11-08
New TensorboardLogger, batch2device fix, FilteringCollateFn, minor refactorings.

hexfaker:


data/collate_fn.py
```diff
@@ -48,3 +51,26 @@ def stack_non_text_features(data, key):
     }
 
     return data
+
+class FilteringCollateFn():
```
Scitator: brackets?

dl/callbacks.py
```diff
@@ -165,6 +167,56 @@ def on_loader_end(self, state):
...
+        logger = SummaryWriter(log_dir)
+        return logger
+
+
+class GroupTensorboardLogger(Callback):
+    # formatter:off
+    def __init__(self,
+            logdir,
+            metric_names: List[str] = None,
+            log_on_batch_end=False
```
Scitator: why don't you add 2 flags? for `batch_end` and `epoch_end`?

dl/state.py
```diff
@@ -9,16 +9,22 @@ class RunnerState(FrozenClass):
     An object that is used to pass internal state during train/valid/infer.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self,
```
Scitator: self to next below pls

Scitator:
@hexfaker Please, check out the CI:
https://travis-ci.com/Scitator/catalyst/jobs/157277820

data/collate_fn.py
hexfaker: Done

dl/callbacks.py
hexfaker: Looks like my solution is more flexible. You can log some metrics on batches, some on epochs, and some on both while using two instances of this callback. Supporting same functionality in one instance would introduce unneeded complexity to code of this callback. Now it's simple and quite readable, I'd prefer to keep it that way.

dl/state.py
hexfaker: done

hexfaker:
@Scitator Yes I did. Most errors produced by lines I even haven't modified. So...

dl/callbacks.py
Scitator: Yup, `You can log some metrics on batches, some on epochs` using only **one** instance of callback. I think, it is much easier to think from metrics perspective. So, log some metric only on batch, some only on epoch, another one – both.

hexfaker: @Scitator Same functionality with more code is not good.

Scitator: Agree, thats why I don't like **two** instances approach.

Scitator:
Thanks! Merged.

PR #22, opened 2018-12-08
Mean average precision callback

lightforever:
Mean average precision callback is implemented.


Scitator:
Hi, could you please merge catalyst:master to your branch?

PR #29, opened 2018-12-21
MetricCallback

TezRomacH:
- Refactor metrics. Now they have to be inherited either from `MetricCallback` or `MultiMetricCallback`
- Remove code duplication
- Apply `yapf` style


utils/metrics.py
```diff
@@ -19,86 +20,136 @@ def precision(output, target, topk=(1,)):
     return res
 
 
-def apk(actual, predicted, k=10):
+def average_precision(outputs, targets, k=10):
     """
     Computes the average precision at k.
-    This function computes the average prescision at k between two lists of
```
Scitator, line 25: precision?

dl/callbacks/metrics.py
```diff
@@ -0,0 +1,185 @@
...
+        self.output_key = output_key
+        self.metric_params = metric_params
+
+    def on_batch_end(self, state):
+        outputs = state.output[self.output_key]
+        targets = state.input[self.input_key]
+
+        metric = self.metric_fn(targets, outputs, **self.metric_params)
+        state.batch_metrics[self.prefix
+                            ] = UtilsFactory.get_val_from_metric(metric)
```
Scitator: codestyle:
```
state.batch_metrics[self.prefix] = \
    UtilsFactory.get_val_from_metric(metric)
```
?

dl/runner.py
```diff
@@ -9,13 +9,13 @@
 
 from catalyst.utils.factory import UtilsFactory
 from catalyst.utils.misc import merge_dicts
-from catalyst.dl.callbacks import Callback, CheckpointCallback, InferCallback
+from catalyst.dl.callbacks.core import Callback, CheckpointCallback, InferCallback
```
Scitator: is it possible to import all callbacks just from `catalyst.dl.callbacks`? maybe, add them to __init__?

losses/dice.py
```diff
@@ -1,35 +1,27 @@
-import torch
 import torch.nn as nn
+from catalyst.utils import metrics
```
Scitator: btw, could we transfer metrics from utils to dl? looks like they finally become DL specific ones
and utils are too complex already

losses/focal_loss.py
```diff
@@ -7,17 +7,18 @@ def __init__(self, gamma=2):
         super().__init__()
         self.gamma = gamma
 
-    def forward(self, input, target):
-        if not (target.size() == input.size()):
+    # TODO refactor
+    def forward(self, outputs, target):
```
Scitator: targets?

utils/metrics.py
```diff
@@ -19,86 +20,136 @@ def precision(output, target, topk=(1,)):
...
+    Returns
+    -------
+    score : double
+            Dice score
+    """
+    if activation is None or activation == "none":
+        activation_fn = lambda x: x
+    elif activation == "sigmoid":
+        activation_fn = torch.nn.Sigmoid()
+    elif activation == "softmax":
```
Scitator: is it better to say softmax2d then? or use `toch.nn.__dict__[name] if name in []`?

TezRomacH:
fixed:
- add `callbacks.core` to `callbacks.__init__` 
- change quotes
- move `metrics` from `catalyst.utils`  to `catalyst.dl`

PR #47, opened 2019-01-09
Registry refactor

TezRomacH:
Usage of `Registry`:

```python
from catalyst.contrib.registry import Registry

@Registry.callback
class MyPerfectCallback(Callback):
    ...
```

or 

```python
from catalyst.contrib.registry import Registry

class MyPerfectCallback(Callback):
    ...

Registry.callback(MyPerfectCallback)
```

also supports `optimizer`, `criterion`, `callback`, and `model`

```python
from catalyst.contrib.registry import Registry

@Registry.optimizer
class UltraFastQuasiNewtonOptimizer:
    ...
```


Scitator:
why have you used `contrib/registry/__init__.py` rather then just `contrib/registry.py`?

TezRomacH:
> why have you used `contrib/registry/__init__.py` rather then just `contrib/registry.py`?

well, didn't think about it. i'll change it

TezRomacH:
I have tested the code on a toy example. And you can be sure, it works

PR #49, opened 2019-01-10
yapf hack to import all callbacks

TezRomacH:
- improved readability 
- all new callbacks will be automatically added to `__init__.py`, it will allow you to use `from catalyst.dl.callbacks import ...`
- no need to change `yapf` config with with hack

### DIFF
--- a/.yapfignore
+++ b/.yapfignore
@@ -1,0 +1,1 @@
+contrib/criterion/__init__.py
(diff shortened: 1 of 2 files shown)


TezRomacH:
wtf, why did the build not compile? During the docs, it should
https://github.com/google/yapf#why-does-yapf-destroy-my-awesome-formatting

🤔

TezRomacH:
bth why `.yapfignore` doesn't work???
https://github.com/google/yapf/blob/master/README.rst#excluding-files-from-formatting-yapfignore

PR #57, opened 2019-01-14
Refactoring01

Scitator:


contrib/registry.py
```diff
@@ -76,3 +66,104 @@ def model(*models_factories: Factory) -> Union[Factory, List[Factory]]:
...
+            return None
+        elif isinstance(name, nn.Module):
+            return name
+        elif isinstance(name, str):
+            return _REGISTERS["module"][name]
+        else:
+            return name
+
+    @staticmethod
+    def create_callback(callback=None, **callback_params):
```
TezRomacH: Maybe rename these methods?
It's a little confusing, that registry has method `callback` and `create_callback`

also for `optimizers`/`model` ...

TezRomacH: i propose `get_callback`

contrib/registry.py
```diff
@@ -76,3 +66,104 @@ def model(*models_factories: Factory) -> Union[Factory, List[Factory]]:
...
+            return None
+
+        criterion = _REGISTERS["criterion"][criterion](**criterion_params)
+        if torch.cuda.is_available():
+            criterion = criterion.cuda()
+        return criterion
+
+    @staticmethod
+    def create_model(config, available_networks=None):
+        model_params = config.pop("model_params", {})
```
TezRomacH: it's bad. Registry has not to know about any `config`

contrib/registry.py
```diff
@@ -76,3 +66,104 @@ def model(*models_factories: Factory) -> Union[Factory, List[Factory]]:
...
+    def create_optimizer(
+            model, fp16=False, optimizer=None, **optimizer_params
+    ):
+        if optimizer is None:
+            return None
+
+        master_params = list(
+            filter(lambda p: p.requires_grad, model.parameters())
+        )
+        if fp16:
```
TezRomacH, line 148: let's create `catalyst.dl.fp16`'s function `def to_fp16` and move all from this if to the methos

TezRomacH (changes requested):

TezRomacH (approved):

PR #71, opened 2019-01-22
fix paths

TezRomacH:


.gitignore
```diff
@@ -109,10 +109,8 @@ ENV/
 tmp/
 logs/
 
-*.ipynb
```
TezRomacH: this one is for to be able to create new examples

.gitignore
```diff
@@ -109,10 +109,8 @@ ENV/
...
 *.csv
 *.tsv
-*.txt
```
TezRomacH: this one for updating `requirements.txt`

.gitignore
Scitator: all current examples already added and tracked by git
for new ones, you can easily use `git add -f examples/new_ex.ipynb`

.gitignore
Scitator: just like with .ipynb, all current requirements already tracked

PR #91, opened 2019-01-27
Add scale support for LRFinder

ivbelkin:


Scitator:
Hi, thanks for contribution!

The only thing to do better is define something like `_calr_lr` during init to remove if-else usage every lr update. You can see an example of such approach in https://github.com/catalyst-team/catalyst/blob/master/catalyst/rl/agents/layers.py#L117

ivbelkin:
Hi, thanks for fast review and hint!

ivbelkin:
Have you any idea why there are two checks that seem identical? Did I something wrong?

Scitator:
Checks are okay. Thanks!
There are only some Travis bugs, so nevermind.

PR #101, opened 2019-02-02
remove logdir hack in runner

ivbelkin:
Add base class for anything that needs 'logdir' to be specified in 'train' mode - LoggerCallback.

'logdir' attribute implicitly added to RunnerState through kwargs (probably, it would be better to add it explicitly).

LoggerCallback becomes base class for CheckpointCallback, Logger, TensorboardLogger.


Scitator (approved):

Scitator:
So, all looks great! Finally, this hack was removed.

Btw, you found nice solution, cause this idea of transferring all experiment config and parameters in one place is one of the key features of 19.02 release. Maybe we should chat somehow :)

PR #103, opened 2019-02-03
Version/19.03

Scitator:
catalyst-dl 19.02 proposal

main goal:
```
from catalyst import Runner
from expdir import MyExperiment

Runner.run(MyExperiment) (train/infer)
```

typical run:
```
mode (train/infer)
	stage
		epoch
			loader
				batch
```
during stage - model/etc are the same, between stages - can be easily replaced

main entities:
- Registry - Factory for registering user extentions
- Experiment - keeper of the config, knows how to create model / etc, but does not keep them
- State - all infos about what is in the experiment now, in the current stage
- Runner - runner, responsible for main logic
- Callbacks - additional user extentions for changing runner’s work a bit


ivbelkin:
Hi!
New experiment-runner architecture looks really general and customizable!
However, there are some thing I want to note.
1. One of the best things of current version is **YAML configs**. Would they be deprecated in future? If no (I hope), what would be their place in ecosystem? Is experiment instance would be created based on parameters in yaml?
2. **Schedulers** Role of SchedulerCallback is clear, but there are also LRUpdater callbacks. Why not to use SchedulerCallback with custom schedulers derived from torch.optim.lr_scheduler._LRScheduler only?
3. I think, there is need to rethink callbacks and events. Why? E.g. combining gradient accumulation in OptimizerCallback and SchedulerCallback lead to changing optimizer parameters during accumulation. The way to cope with this, and, may be, some other unpleasant things caused by calling all callbacks that implement particular event handler method, is add ability to generate events from callbacks. E.g. OptimizerCallback would generate on_weights_update event after accumulation, and SchedulerCallback would handle this event by adjusting optimizer parameters if needed and produce on_optimizer_update event. Something like event pool. And runner, instead running particular events, will add its own events (that are now provided) to pool and run handling them with event added by callbacks.
4. Specifying different learning rates for parameter groups in yaml config. And working with parameter groups at all.

hexfaker:
@ivbelkin 
1. No, of course it will not be removed. YAML configs parser will be just one of `Experiments`.
2. The idea is good, we will think about it. But there is no simple way to do it, because `_LRScheduler` can update lr only at the beginning of epoch, and `LRUpdater` lets us to do it every iteration.
3. There is definitely a problem with matching callbacks between each other. But  introducing events like you suggest leads to huge increase of complexity. Catalyst's learning curve is already steep enough, and we don't want make it even steeper. Personally I think the advanced functionality like gradient accumulation should be removed from upstream to prevent confusions like yours. The ones who need it should implement it themselves.
4. Thanks, will think about it.

Scitator:
So, and now the official answer :)

1. Yaml configs are must have for reproducible research, so they still will be with us. WIP for Yaml-based `Experiment`.
2. Custom schedulers are adapted from `fast.ai`. Introduced in early 2018, they are still not implemented in PyTorch officially. Nevertheless, you are right, contributions on make them PyTorch-friendly are welcome.
3. Interesting idea, but looks too complicated for current needs. Introducing new event - is not the answer, but raises a question of "how make this complex thing easier?". Current callbacks event are quite easy to understand and general. Maybe, we just need to rethink and simplify them.
4. As model/criterion/optimizer are on the users side it's quite hard to standardize. There are `get_something_stuff` methods, that just took specified params from config and output their results to specific `state` field (model, optimizer, etc). So you can easily redefine on of the methods and different learning rates for parameter groups. Looks a bit complicated and yup, maybe we should add some easy-to-use logic for that.

ivbelkin:
Hi! Thanks for answers!

2. @hexfaker, in _LRScheduler they use term `epoch`, but it is nothing but some counter of `step` method callings. So using `SchedulerCallback` with `mode="batch"` will make updates every iteration with _LRScheduler correctly. However, LRUpdater has full access to RunnerState, but is this really needed?

3. Yeah, it complicated. I revealed this concept in Ignite and now I see that it looks really redundant for catalyst architecture. FYI: now in fast.ai they use four more callback events: on_loss_begin, on_backward_begin, on_backward_end, on_step_end.

Scitator:
3. As for me, `fast.ai` callbacks events looks a bit... overhead. Would be grateful if someone can show me any examples when all these `on_loss_begin/on_backward_begin/on_backward_end/on_step_end` really must have.

catalyst/dl/experiments/runner.py
```diff
@@ -0,0 +1,177 @@
...
+        self.state: RunnerState = None
+        self.stage: str = None
+
+        self.callbacks: OrderedDict[str, Callback] = None
+
+        if device is None:
+            self._prepare_model()
+
+    @staticmethod
+    def _batch2device(batch: Mapping[str, Any], device):
```
ferrine: What about this design?
https://github.com/pytorch/pytorch/blob/master/torch/nn/parallel/scatter_gather.py#L11

ferrine:
Also what about saving full config with defaults in here?
https://github.com/catalyst-team/catalyst/blob/master/catalyst/utils/config.py#L119
Say here https://github.com/catalyst-team/catalyst/blob/master/examples/cifar_stages/model.py#L10
you may have something like this:
```
def __init__(self, depth=7):
    ...
```
The default arg is 7, but it can also be defined via config. It is better move this sensitive value to dumped config.
Why? I'm planning to make https://github.com/ferrine/exman and catalyst friends. And this particular feature is important to make pandas know about default values

belskikh:
What about passing lr, num_epochs, optimizer, scheduler to Runner.train() as parameters?

Scitator:
@belskikh I think, because of the stages-support all this params should be in Experiment.

belskikh:
If I understand correctly, Experiment is a some kind of parametrizer of a train() method, and is needed for reproducibility etc? 
If so, why should user create an another abstraction instead of using python's method interface? I think Runner could wrap these params and create an Experiment object to keep stages-support erc.

Scitator:
@belskikh Could you please give more context on your idea?
Nevertheless, if I understood you correctly, you are arguing about Runner vs Experiments abstractions? or why should not we use just one of them?

Personally, I think, that to differ 2 part of experiment run:
- Experiment - **what** your want to run, model, stages, etc
- Runner - **how** you want to run it; for example, Supervised, Unsupervised, Gans, etc

Usually, it looks like you need to rewrite only **what** part of it - Experiment. Cause, runner main logic is quite general and can be easily modified by callbacks mechanism. 
This hypothesis would be tested during examples refactoring :)

Hope this explanation would help, and looking forward to your idea explanation. Some psedo-code, maybe?

PS. just to be sure, you are looking for something like this?
```
from expdir import Experiment
Experiment.train() (train/infer)
```

belskikh:
Yeah, kind of
I can assume that not every Runner type will fit every Experiment params.

I think about another abstraction - Runner is *what* - model, loaders, loss; and Runner.train() is *how* - stages, lr, epochs, optimizer etc.

So, SomeRunner is a pipeline for the Some type of architecture (gans, unsupervised etc) and its `train() ` method provides interfaces for this pipeline customization and running

TezRomacH:
Agreed with @belskikh about a runner. It should know only how to run an experiment and should have parameters like `batch_size`, `n_epochs`, `n_workers`. 

Experiment know only about its callbacks, model, optimizer and criterion. But how to run this callbacks know only Runner `SupervisedRunner`, `GanRunner` etc. Runner can contain multiple experiments for different stages

hexfaker:
No, @belskikh @TezRomacH, can't agree with you. Loss and loaders are usually different in different stages. This means that they _are_ stages themselves. `batch_size` and `n_epochs` have the same value to model as `lr`, so they are _what_, not _how_.

Yes, not all experiments will match all runners, but they don't have to. In most cases the runner you need is already there, and you can just use it. Also, sometimes, there is even not needed to update Experiment. Because `ConfigExperiment` will cover your needs.

belskikh:
I think that `batch_size` and `num_workers ` parameters are usually defined in the DataLoader, that can be created separately from the Runner

belskikh:
@hexfaker I think, that the only models parameter is its weights, that obtained during training process, which is stages, losses, batches, lrs, epochs etc.

hexfaker:
@belskikh What are you talking about? We were not talking about what is model parameter and what is not.

belskikh:
I say, that lr, batch-size and num epochs are *how* for model, not *what*

Scitator:
@belskikh I think, in our case we are thinking about the model as a result of the Experiment. 
And all our abstractions are done around Experiment, rather that model.

From the Experiment perspective, all hyperparameters are just **what** and all run logic is **how**.

belskikh:
Agreed.
But since Experiment is a hyperparameters container, it may be unnecessary level of abstraction, because you can pass these params directly to pipeline

Scitator:
@belskikh 
In simple case, you are correct. Experiment just hold the model, criterion, etc, etc. And all the Runner need to do - take Experiment args and run them all.

Nevertheless, speaking about the config variant (or stages one), where all hyperparameters are in `.yml` now, you need additional abstraction to parse this parameters and feed them to pipeline correctly.

belskikh:
Agreed again.
But what if one want to run pipeline manually? Not from the config, but from the other data source - API etc. Then one should create Runner, Experiment, initialize model, loaders etcetc, but all he need is decrease lr or add/delete callback, or run one more epoch in Jypyter or Kaggle Kernel. 

Can you provide the user with manual interfaces and with optional config pipelines (that are pretty hard to learn, imo) ?

Scitator:
@belskikh Let's be more precise, for now, we have such working example:
```
model = Net()
loaders = {"train": ..., "valid": ....}
exp = BaseExperiment(
    './logs/01',
    model=model,
    epochs=5,
    loaders=loaders,
    callbacks=OrderedDict(
        tqdm=VerboseCallback(),
        loss=LossCallback(),
        optimizer=OptimizerCallback(),
        logger=Logger(),
        tflogger=TensorboardLogger()
    ),
    criterion=CrossEntropyLoss(),
    optimizer=SGD(model.parameters(), lr=0.001, momentum=.9)
)
runner = SupervisedModelRunner(exp, input_key="image")
runner.run("train")
```

I think, that you need to work with model/loaders every time, so there is no problem, that you need to do so.
Decreasing lr, adding scheduler, changing callbacks or epoch num... looks quite easy with such setup, isn't it?

PS. additionally, I think we could rename `batch_handler` to `predict` to make user-friendly inference support. something like
```
batch = {"image": ...}
output = runner.predict(batch)
```

belskikh:
1 line vs 10+ lines, not so easy, imho


Why it is so neccecary, to keep the Experiment? What would happen, if it disappears?

Scitator:
@belskikh Sorry, but I do not see 1 vs 10 line problem and cant understand it. Looking forward to your proposal. It would be much easier to understand each other, if you can also propose solutions.

Rather than saying "not so easy", try to give other people the idea of **why** it is not so easy, **what** would you like to change and **how** do you see it. By this way, this conversation would be much more productive.

belskikh:
I apologize for the possible vagueness of previous statements.

My suggestion is:
- remove Experiment
- add all Experiments parameters as args/kwargs to {SOME}Runner.train method
- for config-dependent situations read config as a dict and pass it as kwargs to  {SOME}Runner.train method

It will allow user to run train manually and reduce amount of code needed for experiments.

Scitator:
@belskikh Okay, and have you any pseudo-code idea of how it should looks like? 
Could your, please, refactor the provided example in your way? So you can make it clear to all our participants.

belskikh:
Here it is:

```model = Net()
loaders = {"train": ..., "valid": ....}
runner = SupervisedModelRunner(input_key="image")

runner.run("train",
    './logs/01',
    model=model, 
    loader=loaders,
    epochs=5,
    callbacks=OrderedDict(
        tqdm=VerboseCallback(),
        loss=LossCallback(),
        optimizer=OptimizerCallback(),
        logger=Logger(),
        tflogger=TensorboardLogger()
    ),
    criterion=CrossEntropyLoss(),
    optimizer=SGD(model.parameters(), lr=0.001, momentum=.9)
)
```

Scitator:
@belskikh Sorry, but....where is the difference?

image

Scitator:
@belskikh Sorry, but... where is the difference?
image


Additionally, I think, that `loaders` should be outside of the `Runner` definition and inside `run` one, that makes your variant even more like first one.

belskikh:
Experiment is redundant, imo, and the difference is - in my example there is no Experiment.

My argument about 10 lines was about a situation, when the user passes a "default" callbacks, optimizer, loss, progressbar, loggers, loaders and model in the Runner class, but I understand, that you doesn't like this approach, and, as I understand, the main goal of this PR is separating "general pipeline" and "what we pass through that pipeline", so these approach have to be verbose, and there is no way to run the same configuration with just , for example, different learning rate without defining the whole set of parameters/loaders/models etc. 
In this way "1 line vs 10+ lines" is not relevant comment, but I still think, that Experiment is redundant in this setup.

Scitator:
So, the final difference of our approaches is just 1 line of code - `exp = Experiment(...)`. In my opinion, 1 line is not so big price for the purpose of different usage availabilities.
This tiny piece of abstraction gives your an opportunity to connect both Jupyter-like Experiments, when you just try bunch of stuff quickly and Config-like Experiments, when you can easily make a grid search in hypterparameters for a week. But it also gives you an opportunity to use stages support easily - the unique feature of catalyst toolset.

Scitator:
Sorry, but you change threads like gloves. I haven't said, that 
> My argument about 10 lines was about a situation, when the user passes a "default" callbacks, optimizer, loss, progressbar, loggers, loaders and model in the Runner class, but I understand, that you doesn't like this approach

Personally, for reproducibility purposes I really like to define the whole experiment in config with all the hyperparameters - it's just much easier to check what was going on there.
Nevertheless, for some specific well-known cases, like classification/segmentation, I think it is possible to predefine some set of callbacks (like loggers, optimizers, etc). But we have to check, if we had such need. 

PS. btw, you can easily make an issue for API simplification for well-known cases. That is always welcome.

belskikh:
I am not an active Catalyst user, and I don't feel that I am should ask you to spend your time on some changes or simplifications or whatever.
I just expressed my thoughts about this approach as you requested in our previous conversation. It seems they were not accepted very warmly, but different opinions are good, in some disputes the truth is born. 
Anyway, I have nothing more to say on topic, so see you later =)

Scitator:
I like our discussion, it gave me a lot of idea to think about. Nevertheless, I usually prefer to see the full context of the conversation topic, rather than mess of different ideas and meanings. And this one was a mess a little bit.
That's why I really like why/what/how approach.

Scitator:
Proposal update: https://gist.github.com/Scitator/0ac3aa517e3be5d8e3c726cc6d413a52
Current working version:
```
loaders = {"train": ..., "valid: ... }
model = Net()
runner = SupervisedRunner(model=model, input_key="image")

# training
runner.train(
    logdir="./logs/01",
    epochs=5,
    loaders=loaders,
    criterion=CrossEntropyLoss(),
    optimizer=SGD(model.parameters(), lr=0.001, momentum=.9),
    callbacks=OrderedDict(
        tqdm=VerboseCallback(),
        loss=LossCallback(),
        optimizer=OptimizerCallback(),
        logger=Logger(),
        tflogger=TensorboardLogger()
    )
)
```

TezRomacH:
My proposal for runner/experiment

UPDATED
https://gist.github.com/TezRomacH/2569a96153373bd0f6d50ef85d5a8eae

hexfaker:
Simple API proposal. This one is separate from the core experiment/runner https://gist.github.com/hexfaker/8b1e0f9aa89905f9424074a1637c39f4

catalyst/dl/experiments/experiment.py
```diff
@@ -187,39 +188,52 @@ def _prepare_stages_config(self, stages_config):
...
     @property
-    def stages(self) -> Iterable[str]:
-        stages_keys = self.stages_config.keys()
+    def stages(self) -> List[str]:
+        stages_keys = list(self.stages_config.keys())
         return stages_keys
 
     def get_total_epochs(self, stage: str):
-        return self.stages_config[stage]["epochs"]
+        return self.stages_config[stage]["state_params"]["total_epochs"]
```
TezRomacH: эх, safitty бы сюды

catalyst/dl/experiments/experiment.py
```diff
@@ -173,16 +165,40 @@ def get_transforms(self, stage: str = None, mode: str = None):
...
-
+        self._config = config.copy()
+        # @TODO: good enough solution?
+        self._config["stages"]["state_params"] = merge_dicts(
+            self._config["stages"].get("state_params", {}),
+            self._config.get("args", {})
+        )
         self.stages_config = self._prepare_stages_config(config["stages"])
+        self._logdir = \
+            self._config.get("args", {}).get("logdir", None) \
```
TezRomacH, line 202: the same result 🙂 
`safe_get(self._config, "args", "logdir")`

ngxbac:
Do we still have command: `catalyst-dl infer ...` 
I do not know how to infer as previous version

Scitator:
@ngxbac Now we have only `catalyst-dl run` for both `train` and `infer` stages.

ngxbac:
> Now we have only catalyst-dl run for both train and infer stages 

How can I do that? I ran this command, but it does not help. Is there any example that I can refer?

```
catalyst-dl run \
   --expdir=$(pwd) \
   --resume=${LOGDIR}/checkpoints/best.pth \
   --out-prefix=${LOGDIR}/dataset.predictions.{suffix}.npy \
   --config=${LOGDIR}/config.json,./inference.yml \
   --verbose

```

Scitator:
@ngxbac You cannot use `expdir` as `pwd` because now your experiment directory should be typical python module.

For examples, you can check out [cifar10 one](https://github.com/catalyst-team/catalyst/tree/master/examples/cifar_stages#local-run) or more advanced - [finetune pipeline](https://github.com/catalyst-team/finetune.catalyst)

ngxbac:
> @ngxbac You cannot use `expdir` as `pwd` because now your experiment directory should be typical python module.
> 
> For examples, you can check out [cifar10 one](https://github.com/catalyst-team/catalyst/tree/master/examples/cifar_stages#local-run) or more advanced - [finetune pipeline](https://github.com/catalyst-team/finetune.catalyst)

Thank you. I checked the examples. I could not do `infer stage` by `catalyst-dl run`. I looked at the code, in my opinion, it seems to me that it does not support `infer` or maybe I have wrong configure/understand.  

I push my repo in [here](https://github.com/ngxbac/datahack-intel-scene) 

I created [manual infer function](https://github.com/ngxbac/datahack-intel-scene/blob/master/inference.py). My observations are:
* if I run `python inference.py` more than two times, it will rise an error `FileExistsError` at [InferenceCallback](https://github.com/catalyst-team/catalyst/blob/master/catalyst/dl/callbacks/inference.py#L31). I think it should be `os.makedirs(os.path.dirname(self.out_prefix), exist_ok=True)`

* `catalyst-dl run` does not support `out-prefix` for `infer stage`, is it? 

I use catalyst for almost projects, competitions anyway. However, I am getting some trouble between inconsistent versions :crying_cat_face:

Scitator:
Inference example is [here](https://github.com/catalyst-team/finetune.catalyst/blob/master/configs/exp_splits.yml#L102) and its run [scripts](https://github.com/catalyst-team/finetune.catalyst/blob/master/bin/run_model.sh).

1. Looks like you are right and we should use `exist_ok=True`. PR would be welcome and thanks 👍 
2. `catalyst-dl run` supports `out-prefix` in some tricky way, let me explain: `out-prefix` was some kind of temporary solution, cause it was too....hardcoded. Now, we found better one, but a little bit more complicated: 
- you can pass any argument you want to `catalyst-dl run`
- these arguments would be transferred to `RunnerState` before run
- in the beginning of run all necessary Callbacks attributes, like `out-prefix` would be taken from `RunnerState` and set to Callback. like [here](https://github.com/catalyst-team/catalyst/blob/master/catalyst/dl/callbacks/inference.py#L23) for example. 
3. Sorry for inconsistent APIs, nevertheless it should have done for final API stabilisation and simplification. Several quite tricky abstractions (like stages) are now much more user-friendly. And `expdir` as python module gives you additional freedom in experiment writing than ever before. All the best for our users :)

ngxbac:
@Scitator  
> Inference example is here and its run scripts. 

I got my way to run infer scripts.

> Sorry for inconsistent APIs, nevertheless it should have done for final API stabilisation and simplification. Several quite tricky abstractions (like stages) are now much more user-friendly. And expdir as python module gives you additional freedom in experiment writing than ever before. All the best for our users :)

I saw a lot of improvements from the upgrade until now. I feel to be more comfortable now. 
Many thanks

PR #104, opened 2019-02-03
Safitty. Proof of Concept

TezRomacH:
Safitty improves readability when you try to take nested keys., for example
```python
list(optimizer.param_groups[0]["params"])
```

can be transformed into
```python
safe_get(optimizer.param_groups, 0, "params")
```

and 
```python
def get_optimizer_momentum(optimizer):
    if isinstance(optimizer, torch.optim.Adam):
        return list(optimizer.param_groups)[0]["betas"][0]	
     elif isinstance(optimizer, torch.optim.SGD)
        return list(optimizer.param_groups)[0]["momentum"]	
    else:	
        return None

```

transforms to

```python
def get_optimizer_momentum(optimizer):
    params = safe_get(optimizer.param_groups, 0)
    return safe_get(params, "betas", 0) or safe_get(params, "momentum")
```

### DIFF
--- a/catalyst/dl/callbacks/base.py
+++ b/catalyst/dl/callbacks/base.py
@@ -91,8 +91,10 @@ class CheckpointCallback(LoggerCallback):
             is_best=is_best,
             is_last=True
         )
-        checkpoint_metric = safe_get(checkpoint, "valid_metrics", main_metric) or \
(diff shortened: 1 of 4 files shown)


Scitator:
moved to 19.03 proposal

PR #111, opened 2019-02-13
fixes

TezRomacH:
- support for not only `--long-keys` but also for shorts `-S`
- if the type is not written explicitly then the parser will use `str` by default (`--key:str` == `--key`)
- fixed error, when slashes at edge produce empty keys. Previously, `--deep/key//=12:int` returns 
```
'deep': {'key': {'': {'': 12}}
```

now it's simply 
```
'deep': {'key': 12}
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ matrix:
 
 stages:
   - Codestyle
+  - Tests
   - Examples
   - Deploy
 
@@ -42,6 +43,13 @@ jobs:
         - flake8 . --count --exit-zero --max-complexity=10 --statistics
         # test to make sure the code is yapf compliant
         - ./yapf.sh --all
+
+    - stage: Tests
+      install:
+        - *requirements
+        - pip install -U pytest
+      script:
+        - pytest
 
     - stage: Examples
       name: "Examples DL"
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,7 @@ jobs:
     - stage: Tests
       install:
         - *requirements
+        - pip install seaborn
         - pip install -U pytest
       script:
         - pytest
--- a/catalyst/utils/config.py
+++ b/catalyst/utils/config.py
@@ -50,8 +50,13 @@ def save_config(config, logdir: str) -> None:
 def parse_config_args(*, config, args, unknown_args):
     for arg in unknown_args:
         arg_name, value = arg.split("=")
-        arg_name = arg_name[2:]
-        value_content, value_type = value.rsplit(":", 1)
+        arg_name = arg_name.lstrip("-").strip('/')
+
+        content = value.rsplit(":", 1)
+        if len(content) == 1:
+            value_content, value_type = content[0], "str"
+        else:
+            value_content, value_type = content
 
         if "/" in arg_name:
             arg_names = arg_name.split("/")
--- a/catalyst/utils/config.py
+++ b/catalyst/utils/config.py
@@ -47,14 +47,6 @@ def save_config(config, logdir: str) -> None:
         json.dump(config, fout, indent=2)
 
 
-def set_arg_to_config(config, key, value) -> None:
-    args_exists_ = config.get("args")
-    if args_exists_ is None:
-        config["args"] = dict()
-
-    config["args"][key] = value
-
-
 def parse_config_args(*, config, args, unknown_args):
     for arg in unknown_args:
         arg_name, value = arg.split("=")
@@ -84,8 +76,13 @@ def parse_config_args(*, config, args, unknown_args):
                 arg_value = eval("%s(%s)" % (value_type, value_content))
             args.__setattr__(arg_name, arg_value)
 
+    args_exists_ = config.get("args")
+    if args_exists_ is None:
+        config["args"] = dict()
+
     for key, value in args._get_kwargs():
-        set_arg_to_config(config, key, value)
+        if value is not None:
+            config["args"][key] = value
 
     return config, args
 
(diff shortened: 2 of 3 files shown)


catalyst/utils/config.py
```diff
@@ -50,8 +50,13 @@ def save_config(config, logdir: str) -> None:
 def parse_config_args(*, config, args, unknown_args):
     for arg in unknown_args:
         arg_name, value = arg.split("=")
-        arg_name = arg_name[2:]
-        value_content, value_type = value.rsplit(":", 1)
+        arg_name = arg_name.lstrip("-").strip('/')
+
+        content = value.rsplit(":", 1)
+        if len(content) == 1:
+            value_content, value_type = content[0], "str"
```
Scitator: no default value, please. can cause undefined behaviour

catalyst/utils/config.py
```diff
@@ -50,8 +50,13 @@ def save_config(config, logdir: str) -> None:
 def parse_config_args(*, config, args, unknown_args):
```
Scitator, line 50: is it okay to finally write tests for this thing?
looks like it is not a DL stuff, so should be deterministic

TezRomacH: yeah sure

catalyst/utils/tests/test_config.py
```diff
@@ -0,0 +1,37 @@
...
+        [
+            "--command", "train", "--path=test.yml:str",
+            "--stages/zero=cero:str"
+        ]
+    )
+
+    configuration, args = config.parse_config_args(
+        config=configuration, args=args, unknown_args=uargs
+    )
+    assert args.command == "train"
```
Scitator, line 34: 🤔 one another suggestion, to move all `args` arguments to config['args'], maybe?

catalyst/utils/tests/test_config.py
```diff
@@ -0,0 +1,37 @@
+# flake8: noqa
+import argparse
+import pytest
+
+from .. import config
+
+
+def test_parse_config_args():
```
Scitator, line 8: and we definitely need to move this thing to travis :)

TezRomacH: done :)

.travis.yml
```diff
@@ -21,6 +21,7 @@ matrix:
 
 stages:
   - Codestyle
+  - Tests
```
Scitator, line 24: ❤️

catalyst/utils/config.py
```diff
@@ -74,6 +75,12 @@ def parse_config_args(*, config, args, unknown_args):
             else:
                 arg_value = eval("%s(%s)" % (value_type, value_content))
             args.__setattr__(arg_name, arg_value)
+
+            args_exists_ = config.get("args")
+            if args_exists_ is None:
+                config["args"] = dict()
+            config["args"][arg_name] = arg_value
+
     return config, args
```
Scitator, line 87: maybe it's better to transfer all `args` arguments to config here? 
just to sync all args with config['args']

some possible draft
```
...

args_exists_ = config.get("args")
if args_exists_ is None:
    config["args"] = dict()
for arg_name, arg_value in args:
    config["args"][arg_name] = arg_value

return config, args
```

catalyst/utils/tests/test_config.py
```diff
@@ -0,0 +1,42 @@
...
+            "--stages/zero=cero:str", "-C=like:str"
+        ]
+    )
+
+    configuration, args = config.parse_config_args(
+        config=configuration, args=args, unknown_args=uargs
+    )
+
+    assert args.command == "train"
+    assert args.path == "test.yml"
```
Scitator, line 35: lets check, that all args kwargs are in configuration["args"]

TezRomacH: done

catalyst/utils/tests/test_config.py
```diff
@@ -0,0 +1,47 @@
...
+    assert configuration.get("stages") is not None
+    assert "zero" in configuration["stages"]
+    assert configuration["stages"]["zero"] == "cero"
+    assert configuration.get("args") is not None
+    assert configuration["args"]["path"] == "test.yml"
+    assert configuration["args"]["C"] == "like"
+    assert configuration["args"]["command"] == "train"
+
+    for key, value in _args._get_kwargs():
```
Scitator: why `_args` rather than `args`?

catalyst/utils/config.py
```diff
@@ -74,6 +86,8 @@ def parse_config_args(*, config, args, unknown_args):
             else:
                 arg_value = eval("%s(%s)" % (value_type, value_content))
             args.__setattr__(arg_name, arg_value)
+            set_arg_to_config(config, arg_name, arg_value)
```
Scitator: why don't you just move
```
for key, value in args._get_kwargs():
    set_arg_to_config(config, key, value)
```
to the end of function, before the `return config, args`?

TezRomacH: updated

catalyst/utils/config.py
```diff
@@ -47,10 +47,19 @@ def save_config(config, logdir: str) -> None:
         json.dump(config, fout, indent=2)
 
 
+def set_arg_to_config(config, key, value) -> None:
```
Scitator: I think, you need to check, that value is not `None`, cause you can corrupt config by parser defaults.

TezRomacH: i think if we pass a value to config it should overwrite it anyway

Scitator: okay, let me explain the full story:
1. you define argparse with some arguments like `--n-workers`, `--batch-size`, etc
2. by default they are None
3. but you have defined them in `config/args`
4. because it's easier to define them in config, rather than pass them every train run

5. you PR feature to transfer all args kwargs to config...you rewrite the config with default args values (None) from parser -> conflict, checks have failed

Scitator: hope fix would help :)

TezRomacH: Thanks, got it

Scitator (approved):

PR #114, opened 2019-02-18
deterministic sac on inference

AlexGrinch:
SAC should produce deterministic actions on inference. Also it seems that some of set_global_seeds calls were superfluous.

### DIFF
--- a/catalyst/rl/agents/layers.py
+++ b/catalyst/rl/agents/layers.py
@@ -177,9 +177,9 @@ class StateNet(nn.Module):
         x = self.head_net(x)
         return x
 
-    def forward(self, *args, with_log_pi=False, **kwargs):
+    def forward(self, *args, with_log_pi=False, deterministic=False, **kwargs):
         x = self._forward_fn(*args, **kwargs)
-        x = self._policy_fn(x, with_log_pi)
+        x = self._policy_fn(x, with_log_pi, deterministic)
         return x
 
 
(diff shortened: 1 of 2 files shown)


catalyst/rl/offpolicy/sampler.py
```diff
@@ -276,7 +276,6 @@ def run(self):
         seed = random.randrange(SEED_RANGE) \
             if self.seeds is None \
             else random.choice(self.seeds)
-        set_global_seeds(seed)
```
Scitator: why have you removed seed specification for new episode?

catalyst/rl/agents/layers.py
```diff
@@ -177,9 +177,9 @@ def _forward_lama(self, observation):
         x = self.head_net(x)
         return x
 
-    def forward(self, *args, with_log_pi=False, **kwargs):
+    def forward(self, *args, with_log_pi=False, deterministic=False, **kwargs):
```
Scitator, line 180: looks like we need to think about better forward API for agents... in future :)

catalyst/rl/offpolicy/sampler.py
```diff
@@ -415,7 +414,6 @@ def run(self):
                     seed = random.randrange(SEED_RANGE)
             else:
                 seed = random.choice(self.seeds)
-            set_global_seeds(seed)
```
Scitator: and here?

AlexGrinch: Returned for further investigation

catalyst/rl/agents/layers.py
AlexGrinch: definitely

PR #117, opened 2019-02-24
Refactor legacy

TezRomacH:


requirements.txt
```diff
@@ -11,6 +11,7 @@ torchnet
 
 # Used in scripts
 matplotlib
+safitty
```
TezRomacH: cat

TezRomacH: i use it as config reader

Scitator: let's discuss safitty integration in future 19.03/04 release
we need to better understand its purpose

TezRomacH: Well, I remember that the main function of Safitty such as `safe_get` and `safe_set` will be discussed. 

But for now, I use only `safitty.load_config(<path_co_config>, ordered=True)` with can read YAML or JSON by giving path. I found it useful because now everything is not clear: in some cases, we use `.yml` configs, but for another (like `tag2class`) we use  `.json`.

This method helps to avoid hardcoded types of markups.

Scitator: since I know, where were taken the main parts of this func, it maybe be better, to return such functionality in the catalyst again :)
and update our `parse` funcs

catalyst/contrib/scripts/split_dataframe.py
```diff
@@ -0,0 +1,122 @@
...
+        dest="in_csv",
+        help="Path to the csv to split",
+        required=True
+    )
+
+    parser.add_argument(
+        "-t", "--train-folds",
+        type=str,
+        dest="train_folds",
+        help="Integers separated by commas. This is represents train folds",
```
Scitator: grammar :)

requirements.txt
TezRomacH: Not really. Catalyst has function `load_ordered_yaml`, but it reads values only from YAML.
On the other hand, there is a `load_config` func in Safitty which is independent to file extension (json or yaml)

Scitator: let's see,
your code:
```python
with config_path.open(encoding=encoding) as stream:
        ext = config_path.suffix

        if ext == ".json":
            object_pairs_hook = OrderedDict if ordered else None
            file = "\n".join(stream.readlines())
            if file != "":
                config = json.loads(file, object_pairs_hook=object_pairs_hook)

        elif ext == ".yml":
            loader = OrderedLoader if ordered else yaml.Loader
            config = yaml.load(stream, loader)
        else:
            raise Exception(f"Unknown file format '{ext}'")
```

part of catalyst `parse_args_uargs` code:
```python
        with open(config_path, "r") as fin:
            if config_path.endswith("json"):
                config_ = json.load(fin, object_pairs_hook=OrderedDict)
            elif config_path.endswith("yml"):
                config_ = load_ordered_yaml(fin)
            else:
                raise Exception("Unknown file format")
```


so, it looks like you can easily refactor `parse_args_uargs` and add some kind of `load_config` to catalyst. especially, when it already has `save_config` functionality

Scitator: later version

Scitator (changes requested):

requirements.txt
Scitator: I really dont like the idea of having dependency on another library only because of one function, which was...borrowed originally from catalyst one. it's weird

Scitator: if you would like to improve functionality - refactoring PR is welcome, but not in such way.

PR #121, opened 2019-03-15
[RL] Global refactoring 2.0, discrete & continuous in the wild

AlexGrinch:
Working version of discrete control, tested on LunarLander-v2.

Things to do before merge:
1. Think how to improve discrete control pipeline to avoid multiple code repetitions.
2. Add support of Atari games and test on Pong.
3. Categorical and quantile support.


catalyst/rl/environments/gym_wrapper.py
```diff
@@ -46,3 +46,47 @@ def step(self, action):
         info["reward_origin"] = reward
         reward *= self.reward_scale
         return observation, reward, done, info
+
+
+class GymWrapperDiscrete:
```
Scitator: why should we have new wrapper? looks like only `observation_shape` and `num_actions` differs.
Is it okay to define them based on their classes? or make some general function to get them from any gym env, cause they are quite standard.

Scitator: btw, we also can introduce some kind of env specification like `type = 'discrete' or 'continuous'`

catalyst/rl/offpolicy/discrete_sampler.py
```diff
@@ -0,0 +1,256 @@
...
+        }
+    state_dict = {key: value.clone() for key, value in state_dict.items()}
+    return state_dict
+
+
+def set_critic_weights(critic, weights, strict=True):
+    critic.load_state_dict(weights, strict=strict)
+
+
+class Sampler:
```
Scitator: to reduce copy-paste I suggest you to specify noise, action and buffer creation functions during the init based on env.type (`'discrete' or 'continuous'`), just like the trainer losses done.

catalyst/rl/offpolicy/scripts/run_discrete_samplers.py
```diff
@@ -0,0 +1,216 @@
...
+from pprint import pprint
+import multiprocessing as mp
+from redis import StrictRedis
+import torch
+
+from catalyst.dl.scripts.utils import prepare_modules
+from catalyst.contrib.registry import Registry
+from catalyst.utils.config import parse_args_uargs, save_config
+from catalyst.utils.misc import set_global_seeds, boolean_flag
+from catalyst.rl.offpolicy.discrete_sampler import Sampler
```
Scitator: if the sampler would be general, there is no need for this additional copy-paste

catalyst/rl/offpolicy/scripts/run_trainer.py
```diff
@@ -51,8 +51,6 @@ def main(args, unknown_args):
         # come on, just refactor whole rl
         environment_fn = Registry.get_fn("environment", args.environment)
         env = environment_fn(**config["env"])
-        config["shared"]["observation_size"] = env.observation_shape[0]
```
Scitator, line 56: so...why have you delete them? (it's a legacy, nevertheless, I'm not sure if we could delete them easily)

catalyst/rl/offpolicy/trainer.py
```diff
@@ -340,13 +222,14 @@ def train(self):
             start_time = time.time()
 
     def update_target_weights(self, step_index):
-        if step_index % self.actor_update_period == 0:
-            self.algorithm.target_actor_update()
-            self.actor_updates += 1
+        if not self.discrete_actions:
```
Scitator: `not self.discrete_actions` sounds a bit confusing, it's more about `actor-critic` logic then.

Scitator (changes requested):

catalyst/rl/offpolicy/discrete_sampler.py
AlexGrinch: Agree

catalyst/rl/offpolicy/scripts/run_trainer.py
AlexGrinch: I deleted it to get compatibility with discrete trainer as I renamed action_size to num_actions in this case to make it more intuitive as it has different meaning.

catalyst/rl/offpolicy/trainer.py
AlexGrinch: For now it seems to me that we need a flag which indicates whether the action space is continuous or discrete. Probably, we can include it inside the environment wrapper as you suggested before.

catalyst/rl/environments/gym_wrapper.py
AlexGrinch: addressed in the last commit

catalyst/rl/offpolicy/scripts/run_discrete_samplers.py
AlexGrinch: run_discrete_samplers.py was deprecated in the last commit

AlexGrinch:
1. Exploration config has been refactored. Now we have the Explorator instance which gathers arbitrary number of exploration strategies and creates one hybrid exploration scheme which selects the strategy to pursue at the beginning of the episode.
2. discrete_sampler and run_discrete_samplers are deprecated.

TODO:
1. Refactor the code and make it solve continuous and discrete versions of Lunar Lander. Currently, it neither crushes nor converges.
2. Add support of Atari games and train the agent to play Pong.

PR #122, opened 2019-03-17
Registery refactoring 

hexfaker:
Big Registry refactoring.

- Separate registry object for every type
- Import-cycle aware
- Common contrib registry, separate dl and rl parts
- Everything refactored and tested

### DIFF
--- a/catalyst/dl/experiments/experiment.py
+++ b/catalyst/dl/experiments/experiment.py
@@ -246,10 +246,10 @@ class ConfigExperiment(Experiment):
         return model
 
     def get_model(self, stage: str) -> _Model:
-        config = self._config["model_params"]
-        fp16 = config.pop("fp16", False)
+        model_params = self._config["model_params"]
+        fp16 = model_params.pop("fp16", False)
 
-        model = MODELS.get_from_params(**config)
+        model = MODELS.get_from_params(**model_params)
 
         if fp16:
             utils.assert_fp16_available()
@@ -283,10 +283,13 @@ class ConfigExperiment(Experiment):
         return optimizer
 
     def get_scheduler(self, stage: str, optimizer) -> _Scheduler:
-        config = \
+        scheduler_params = \
             self.stages_config[stage].get("scheduler_params", {})
 
-        scheduler = SCHEDULERS.get_from_params(**config, optimizer=optimizer)
+        scheduler = SCHEDULERS.get_from_params(
+            **scheduler_params,
+            optimizer=optimizer
+        )
         return scheduler
 
     def get_loaders(self, stage: str) -> "OrderedDict[str, DataLoader]":
@@ -327,8 +330,8 @@ class ConfigExperiment(Experiment):
             self.stages_config[stage].get("callbacks_params", {}))
 
         callbacks = []
-        for key, config in callbacks_params.items():
-            callback = CALLBACKS.get_from_params(**config)
+        for key, callback_params in callbacks_params.items():
+            callback = CALLBACKS.get_from_params(**callback_params)
             callbacks.append(callback)
 
         return callbacks
--- a/catalyst/dl/experiments/experiment.py
+++ b/catalyst/dl/experiments/experiment.py
@@ -1,17 +1,19 @@
-from abc import abstractmethod, ABC
-from typing import Iterable, Any, Mapping, Dict, List
 from collections import OrderedDict
 
 import torch
+from abc import abstractmethod, ABC
 from torch import nn, optim
 from torch.utils.data import DataLoader, Dataset  # noqa F401
+from typing import Iterable, Any, Mapping, Dict, List
 
-from catalyst.contrib.registry import Registry
+from catalyst.dl.registry import \
+    MODELS, CRITERIONS, OPTIMIZERS, SCHEDULERS, CALLBACKS
+from catalyst.dl import utils
 from catalyst.dl.callbacks import Callback  # noqa F401
-from catalyst.dl.callbacks import LossCallback, OptimizerCallback, \
-    SchedulerCallback, CheckpointCallback
-from catalyst.dl.utils import UtilsFactory
+from catalyst.dl.callbacks import \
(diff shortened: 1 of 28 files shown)


catalyst/dl/utils.py
```diff
@@ -11,7 +12,30 @@
 from catalyst.dl.fp16 import Fp16Wrap
 
 
+def pregare_optimizer_params(params, fp16=False):
```
hexfaker: typo

catalyst/dl/utils.py
```diff
@@ -11,7 +12,30 @@
...
+    return master_params
+
+
+def assert_fp16_available():
+    assert torch.backends.cudnn.enabled, \
+        "fp16 mode requires cudnn backend to be enabled."
+
+
 class UtilsFactory:
+    pregare_optimizer_params = pregare_optimizer_params
```
hexfaker: typo

catalyst/rl/offpolicy/algorithms/core.py
```diff
@@ -1,7 +1,9 @@
 import copy
 import torch
 
+from catalyst.dl.registeries import CRITERIONS
```
hexfaker: cycle imports

catalyst/dl/experiments/experiment.py
```diff
@@ -242,30 +245,48 @@ def _postprocess_model_for_stage(self, stage: str, model: _Model):
...
 
     def get_scheduler(self, stage: str, optimizer) -> _Scheduler:
-        scheduler_params = (
-            self.stages_config[stage].get("scheduler_params", {}))
-        scheduler = Registry.get_scheduler(optimizer, **scheduler_params)
+        config = \
+            self.stages_config[stage].get("scheduler_params", {})
+
+        scheduler = SCHEDULERS.get_from_config(
+            "scheduler", config, optimizer=optimizer
```
hexfaker: Hmm, should we move `name_key` inside class and specify it in constructor?

catalyst/rl/offpolicy/algorithms/core.py
```diff
@@ -220,23 +224,23 @@ def prepare_for_trainer(cls, config):
...
-        trainer_action_shape = (config_["shared"]["action_size"], )
+        trainer_state_shape = (config_["shared"]["observation_size"],)
+        trainer_action_shape = (config_["shared"]["action_size"],)
 
-        actor_fn = config_["actor"].pop("agent", None)
-        actor = Registry.get_agent(
-            agent=actor_fn,
+        actror_conf = config_["actor"]
+        actor = AGENTS.get_from_config(
+            "agent",
```
TezRomacH: i don't like the idea to duplicate registry name. If we already specified `AGENTS` it's strange to put another key `"agent"`. Also for any other: `CRITERIONS.get_from_config("criterion", ...`, `CALLBACKS.get_from_config("callback", value)`

Scitator:
So, the PR looks great...but here are some tricks to discuss:

1. Old `Registry.name2nn` was with working with both strings and classes - there was logic similar to yours `_class_or_obj`. The new one `registry.get` works only string and have no checks about classes.
For example, now this [code](https://github.com/catalyst-team/catalyst/blob/2504de7ad7e2e85327ba9544d6ca8c6ff1536e06/catalyst/rl/agents/critic.py#L24) will not work: `activation_fn` is not an string there.
So, as for me, it's better to integrate such checking into new `registry`....we can call it `get_(***)` to emphasize some logic under the hood.

2. Secondly, you create [`registry.get_from_config`](https://github.com/catalyst-team/catalyst/blob/2504de7ad7e2e85327ba9544d6ca8c6ff1536e06/catalyst/utils/registry.py#L176) to create instances of classes right from the registry. But this particular method looks too complicated and raises many questions. For example,
- why do we need `name_key` ? I have not seen any usage of this feature. I see some additional flexibility with it.... and even more misunderstanding of it....it just too tricky and looks like premature optimization :)
- additionally, why have you support both `config` and `**kwargs`? you also do `config.update(kwargs)` in the end...should not we leave only one then?
So, my suggestion here is to... simplify and refactor a little bit:
```python
some code was here
```
During the example coding, I finally understand, that it is much simpler to remove `get_from_config` and use `get_instance` instead.
`get_from_config` variant:
```python
abc = registry.get_from_config(
    {"name": "my_fn", "args1": 1, "args2": 2},
    args3=3
)
```
 
`get_instance` variant:
```python
abc = registry.get_instance(
    name="my_fn", 
    args1=1, 
    args2=2,
    args3=3
)
```
As for me, the second variant looks simpler and more readable. And it does not force you to pass dictionary variable where they are not necessary.

hexfaker:
1. Actually I thought about this, but I was'n sure this use case is common enough. Perhaps it is.
2.  `get_from_confg` & `get_instance`
2.1 Actually `name_key` is not an additional flexibility. When this method was first introduced, I was sure there will be cases in our existing codebase where config dict will have not standard key for name. I actually was wrong. It's okey to remove this particular parameter.
2.2 How do you plan to use `get_instance` in [this](https://github.com/catalyst-team/catalyst/blob/2504de7ad7e2e85327ba9544d6ca8c6ff1536e06/catalyst/dl/experiments/experiment.py#L263) case for example? Or you are suggesting to merge functionality of this methods?  In this case, I'd like to leave them as is because code is more readable this way.

Scitator:
Okay, I remembered our lovely constructions with predefined "names"
```
scheduler_params:
    scheduler: MultiStepLR
    milestones: [10]
    gamma: 0.3
```

Nevertheless, I still think, that such kind of simplification would be nice:
```python
def get_from_params(
    self,
    meta_factory=None,
    **kwargs
) -> Union[Any, Tuple[Any, Mapping[str, Any]]]:
    params = copy.deepcopy(kwargs)
    name = params.pop(self._name_key)
    return self.get_instance(name, meta_factory=meta_factory, **params)
```

So the critetion example, should be solved with something like:
```python
criterion_params = self.stages_config[stage].get("criterion_params", {})
criterion = CRITERIONS.get_from_params(**criterion_params)
```

Scitator (approved):

PR #123, opened 2019-03-17
`os.makedirs(..., exist_ok=True)` on inference

TezRomacH:


catalyst/dl/callbacks/inference.py
```diff
@@ -28,7 +28,7 @@ def on_stage_start(self, state):
         if self.out_dir is not None:
             self.out_prefix = str(self.out_dir) + "/" + str(self.out_prefix)
         if self.out_prefix is not None:
-            os.makedirs(os.path.dirname(self.out_prefix))
+            os.makedirs(os.path.dirname(self.out_prefix), exist_ok=True)
```
TezRomacH, line 31: the main change

Scitator (approved):

catalyst/dl/utils.py
```diff
@@ -53,7 +53,7 @@ def create_tflogger(logdir: str, name: str) -> SummaryWriter:
     @staticmethod
     def create_loggers(
         logdir: str, loaders: List[str]
-    ) -> 'OrderedDict[str, SummaryWriter]':
```
TezRomacH: i think, for types is better to put `'` quotes to note, that it is not a typical string

PR #138, opened 2019-04-07
registry_as_mapping

TezRomacH:
Implement python's mapping methods for Registry class

Examples:
```python
>>> from catalyst.dl import registry

>>> "LinkNet" in registry.MODELS()
True

>>> registry.MODELS().all()
['UNet', 'ResNetUnet', 'LinkNet']

>>> [name for name in registry.CALLBACKS if "LR" in name]
['LRUpdater', 'OneCycleLR', 'LRFinder']

>>> len(registry.CALLBACKS)
71
```

### DIFF
--- a/catalyst/rl/random_process.py
+++ b/catalyst/rl/random_process.py
@@ -82,11 +82,10 @@ class OrnsteinUhlenbeckProcess(AnnealedGaussianProcess):
         self.reset_states()
 
     def sample(self):
-        x = (
-            self.x_prev + self.theta * (self.mu - self.x_prev) * self.dt +
-            self.current_sigma * np.sqrt(self.dt
-                                         ) * np.random.normal(size=self.size)
-        )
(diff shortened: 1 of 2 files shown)


TezRomacH:
yapf is arguing to RL module

catalyst/rl/random_process.py
```diff
@@ -82,11 +82,10 @@ def __init__(
         self.reset_states()
 
     def sample(self):
-        x = (
-            self.x_prev + self.theta * (self.mu - self.x_prev) * self.dt +
-            self.current_sigma * np.sqrt(self.dt
-                                         ) * np.random.normal(size=self.size)
-        )
+        theta = self.theta * (self.mu - self.x_prev)
```
TezRomacH, line 85: only for `yapf`

Scitator (approved):

PR #148, opened 2019-04-19
catalyst.core + catalyst.classification = now kiss

Scitator:
- epoch + stages fix
- AUC callback
- Confusion Matrix Callback
- FocalLoss
- MiniEpochSampler


TezRomacH:
docstrings for new Callbacks + add to docs file

PR #156, opened 2019-04-25
Fix stratified split df fold column type from float to int

gazay:
When checked https://github.com/catalyst-team/classification locally - noticed that fold column is float which is not a big deal but have no sense. Decided to fix and try to provide small test for it.

I really like your library and code style and organization and I think I can help here and there sometime with some tests and improvements. Do you have any testing conventions?

### DIFF
--- a/catalyst/utils/data.py
+++ b/catalyst/utils/data.py
@@ -213,7 +213,7 @@ def stratified_fold_split(
     skf = StratifiedKFold(
         n_splits=n_folds, shuffle=True, random_state=random_state
     )
-    fold_column = np.zeros(len(dataframe))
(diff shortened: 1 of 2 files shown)


gazay:
I've done this data preparation step so complex because I was thinking to test other aspects of this method (e.g., working with method attributes, check number of folds, etc.). But then I hesitated because I'm not sure about your testing conventions - should I move data preparation step in fixture, should I test everything in one function or should split test objects and so on.

gazay:
I've added few more tests and move data setup to extra method.

TezRomacH (approved):
Thanks! Your tests are very helpful ❤️

PR #161, opened 2019-04-26
Refactor test_tag2label with tmp_path usage

gazay:
I planned to continue writing tests for utils/data but its `create_dataset` method brought me to this test. I think that dummy data should be, if possible, encapsulated -- so here I propose to use `tmp_path` fixture which according to pytest docs will provide unique tmp dir path for every test case. Also I tried to simplify setup code and removed excess library imports.

### DIFF
--- a/catalyst/contrib/scripts/tests/test_tag2label.py
+++ b/catalyst/contrib/scripts/tests/test_tag2label.py
@@ -2,14 +2,14 @@ from ..tag2label import prepare_df_from_dirs
 
 
 def _setup_dataset_fs(tmp_path):
-    def create_children(p, children):
-        for c, sub_children in children.items():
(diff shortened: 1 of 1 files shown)


TezRomacH (changes requested):
Hi! Thanks for the PR!

Could you please change the variables' names to our code style? I mean, change abbreviations like `tmp` and one-letter variables to the full form

gazay:
@TezRomacH I've changed variable names in helper methods to full form, although I don't think that `tmp_path` should be changed, because it's standard pytest fixture (https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture)

TezRomacH (approved):

PR #162, opened 2019-04-28
onecycle scheduler

TezRomacH:
- `OneCycleLR` scheduler with warmup stage and linear annealing for `LR` and `Momentum`
- fixed `get_optimizer_momentum`. For now, it can work not only with Adam and SGD, but for any others Pytorch optimizers
- base safitty support (according to https://github.com/catalyst-team/catalyst/projects/4#card-18173195)

TODO:
- [x] docs
- [ ] more complex annealings like `log` or `cosine` (nex release)


catalyst/contrib/scheduler/onecycle.py
```diff
@@ -0,0 +1,67 @@
...
+
+class OneCycleLR(_LRScheduler):
+    def __init__(
+            self,
+            optimizer: Optimizer,
+            num_epochs: int,
+            lr_range=(1.0, 0.005),
+            init_lr: float = None,
+            warmup_epochs: int = 0,
+            decay_epochs: int = 0,
```
Scitator: I think, it's better to use `warmup_factor` and `decay_factor` (not sure about the factor suffix, btw, we need something better) and produce `warmup_epochs  = num_epochs  * warmup_factor`, `decay_epochs  = num_epochs * decay_factor`

nevertheless, I understand, that we can keep both and use one, that is not None
and do not forget, that this scheduler should work not only in `epoch` mode, but in the `batch` one too (cause the main feature of 1Cycle is lr & momentum change during the epoch, after each batch)

PS. just for code beauty it is much better to group variables by their meaning, something like
```
optimizer: Optimizer,
num_epochs: int,  # epochs block
warmup_epochs: int = 0,
decay_epochs: int = 0,
lr_range=(1.0, 0.005),   # lr block
init_lr: float = None,
momentum_range=(0.8, 0.99, 0.999),   # momentum block
init_momentum: float = None,
```

catalyst/contrib/scheduler/onecycle.py
```diff
@@ -0,0 +1,218 @@
...
+from catalyst.dl.utils import get_optimizer_momentum
+from .base import BaseScheduler
+
+
+class OneCycleLR(BaseScheduler):
+    """
+    OneCycle scheduler with warm-up & lr decay stages.
+    First stage increases lr from ``init_lr`` to ``max_lr``,
+    and called ``warmup``. Also it decreases momentum
+    from ``init_momentum`` to ``min_momentum``. Takes ``warmup_epochs`` epochs
```
Scitator: epochs?

catalyst/contrib/scheduler/onecycle.py
```diff
@@ -0,0 +1,218 @@
...
+            init_momentum, min_momentum, max_momentum, final_momentum
+
+        self._calculate_lr_momentum(
+            warmup_steps,
+            lr_annealing_steps,
+            decay_steps
+        )
+
+        self.total_groups = len(optimizer.param_groups)
+        super().__init__(optimizer)
```
Scitator, line 97: why super in the end?

catalyst/contrib/scheduler/onecycle.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def _calculate_warmup(
+            self,
+            num_steps: int,
+            warmup_steps: int,
+            warmup_fraction: float
```
Scitator, line 103: maybe, make warmup_steps & warmup_fraction None by default and check that one of the is not None?

catalyst/contrib/scheduler/onecycle.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self.warmup_steps = warmup_steps
+        self.has_warmup = warmup_steps != 0
+        return self.warmup_steps
+
+    def _calculate_decay(
+            self,
+            num_steps: int,
+            decay_steps: int,
+            decay_fraction: float
```
Scitator, line 119: same as previous comment

catalyst/contrib/scheduler/onecycle.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self.decay_steps = decay_steps
+        self.has_decay = decay_steps != 0
+        return self.decay_steps
+
+    def _calculate_lr_momentum(
+            self,
+            warmup_steps: int,
+            lr_annealing_steps: int,
+            decay_steps: int
```
Scitator, line 135: same as previous comment

catalyst/contrib/scheduler/onecycle.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+
+        lr_annealing_steps = num_steps - (warmup_steps + decay_steps)
+
+        self.warmup_steps = warmup_steps
+        self.lr_annealing_steps = lr_annealing_steps
+        self.decay_steps = decay_steps
+        self.num_steps = warmup_steps + lr_annealing_steps + decay_steps
+
+        self.lr_range = init_lr, max_lr, min_lr, final_lr
```
Scitator, line 86: why not keep them all as attributes?

catalyst/dl/callbacks/base.py
```diff
@@ -268,26 +270,43 @@ def __init__(
...
 
+        if isinstance(scheduler, OneCycleLR) and self.mode == "batch":
+            self._onecycle_recalculated = False
+            scheduler.reset()
+
+    def on_loader_start(self, state: RunnerState):
+        scheduler = state.get_key(
+            key="scheduler", inner_key=self.scheduler_key
+        )
+        if state.loader_name == "train" and \
```
Scitator: it's better to use `state.loader_name.starswith("train")` or `state.need_backward`

catalyst/dl/callbacks/base.py
```diff
@@ -268,26 +270,43 @@ def __init__(
...
+                isinstance(scheduler, OneCycleLR) and self.mode == "batch":
+            if not self._onecycle_recalculated:
+                scheduler.recalculate_(loader_len=state.loader_len)
+                self._onecycle_recalculated = True
```
Scitator: looks like it can work incorrectly with more than 1 train loaders, could you please check it with several of them?

catalyst/dl/callbacks/base.py
```diff
@@ -296,6 +315,22 @@ def on_epoch_end(self, state):
         if self.mode == "epoch":
             self.step(state=state)
 
+    def _scheduler_step(
```
Scitator, line 323: make it static or outer function?

catalyst/dl/utils.py
```diff
@@ -214,3 +216,37 @@ def get_activation_by_name(activation: str = None):
             "Activation implemented for sigmoid and softmax2d"
         )
     return activation_fn
+
+
+def get_optimizer_momentum(optimizer: Optimizer) -> float:
```
Scitator, line 221: move scheduler step here?

Scitator (approved):

PR #181, opened 2019-05-10
load optimizer from previous stage

bagxi:


catalyst/dl/experiments/experiment.py
```diff
@@ -308,10 +307,22 @@ def _get_optimizer(*, model_params, **params):
                 optimizer[key] = ConfigExperiment._get_optimizer(
```
Scitator: as `_get_optimizer` is now not static, should we use `self. _get_optimizer` instead?

Scitator (approved):

PR #183, opened 2019-05-13
Apex support for mixed precision training

BloodAxe:
Makes mixed precision training as simple as possible:

```
model, optimizer = amp.initialize(model, optimizer, opt_level="O1")

runner = SupervisedRunner()
runner.train(
    model=model,
    criterion=criterion,
    optimizer=optimizer,
    scheduler=scheduler,
    ...)
```

Breaking changes (mostly internal, but still):
1. `Experiment.get_optimizer` renamed to `Experiment.get_optimizer_and_model` and returns optimizer and model as a tuple. Motivation: Apex requires two objects to initialize - a model and optimizer. A model must be already on GPU and not wrapped with DataParallel/DistributedDataParallel. This change was made to have clear amp initialization in one place.
2. `UtilsFactory.prepare_model` now does not wrap model with DataParallel since this will break Apex initialization. Instead, DataParallel applied in `Experiment.get_optimizer_and_model` if fp16 is not enabled and more than one GPU is available.


Scitator:
@BloodAxe proposal: add fp16 flag to `SupervisedRunner` `.train` and `.infer` with automatic model cast to fp16 mode. What do you think about it?

BloodAxe:
This need discussion. Apart from global enable/disable fp16 training mode flag, there are couple of flags that user may want to change (optimization level, flag to keep batch-norm in fp32, loss scaling flags, etc).
This would expand set of parameters in `SupervisedRunner`. That's my only objection so far.

catalyst/dl/experiments/experiment.py
```diff
@@ -126,7 +125,6 @@ def __init__(
         self._minimize_metric = minimize_metric
         self._verbose = verbose
         self._additional_state_kwargs = state_kwargs or {}
-        self.checkpoint_data = checkpoint_data or {}
```
Scitator: somehow you have removed your own `checkpoint_data` feature during merge :)

Scitator:
@BloodAxe hmm, you are right...
then, what about fp16 flag and the following logic:
- do nothing if model & optimizer are already fp16
- turn default optimization level (01?), keep batch norm in fp32 and loss autoscaling (or scale by 128)
?

.travis.yml
```diff
@@ -49,13 +49,17 @@ jobs:
         - *requirements
         - pip install seaborn
         - pip install -U pytest
+      before_script:
```
Scitator: do we need apex, if there are no fp16 tests?

BloodAxe:
@Scitator I think it's still make sense to have fast&quick option to turn on fp16 mode for SupervisedRunner.
I'll fix remaining issues and add `fp16` flag in a while.

PR #184, opened 2019-05-13
Added extra checkpoint data

BloodAxe:
Allows to save additional metadata to checkpoint files:

```
runner.train(
        model=model,
        ....,
        extra_checkpoint_data={
            'class_names': ['Cat', 'Dogs']
        }
)
```

### DIFF
--- a/catalyst/dl/callbacks/base.py
+++ b/catalyst/dl/callbacks/base.py
@@ -111,7 +111,8 @@ class CheckpointCallback(Callback):
             valid_metrics=dict(state.metrics.valid_values),
             stage=state.stage,
             epoch=state.epoch,
-            extra_checkpoint_data=state.extra_checkpoint_data
+            extra_checkpoint_data=state.extra_checkpoint_data \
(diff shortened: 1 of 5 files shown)


catalyst/dl/callbacks/base.py
```diff
@@ -102,6 +102,9 @@ def on_epoch_end(self, state: RunnerState):
         if state.stage.startswith("infer"):
             return
 
+        extra_checkpoint_data = state.extra_checkpoint_data \
+            if hasattr(state, 'extra_checkpoint_data') else {}
```
Scitator: let's just add `checkpoint_data` to state init, so we can use it also from Config API.
proposed way:
```
state_params:
    checkpoint_data:
        exp: "thin is my best experiments"
        date: "today"
```

This will also reduce such `hasattr` check.

Scitator (approved):

hexfaker (changes requested):
Let's squash all small commits into one or two. Keep our history simple and clean!

PR #189, opened 2019-05-20
Catalyst model tracing preview

### LINKED ISSUE #180: [feature] add jit.trace support
Add automated model tracing & packing.

hexfaker:
Fixes #180


catalyst/dl/scripts/trace.py
```diff
@@ -0,0 +1,125 @@
...
+    Model that calls specified method instead of forward
+
+    (Workaround, sigle method tracing is not supported)
+    """
+    def __init__(self, model, method_name):
+        super().__init__()
+        self.model = model
+        self.method = method_name
+
+    def forward(self, *input):
```
Scitator: what about `*args, **kwargs` usage?

catalyst/dl/experiments/experiment.py
```diff
@@ -70,13 +69,18 @@ def get_model_stuff(self, model, stage: str):
     def get_callbacks(self, stage: str) -> "List[Callback]":
         pass
 
+    @abstractmethod
```
Scitator: all we need during training is loaders, so I am not sure it is good idea to force user to implement `get_datasets` and `get_datasets_only`

catalyst/dl/scripts/trace.py
```diff
@@ -0,0 +1,125 @@
...
+    model.load_state_dict(torch.load(checkpoint_path)["model_state_dict"])
+    model.eval()
+
+    for p in model.parameters():
+        p.requires_grad_(False)
+
+    tracer = _TracingModelWrapper(model, method_name)
+    runner: Runner = runner_type(tracer.cpu(), torch.device("cpu"))
+
+    train_dataset = experiment.get_datasets_only(stage)["train"]
```
Scitator: we do not know datasets names, so I suggest use something like
```
datasets.key()[0]
```

Scitator: and what about get loaders, take first one and get dataset from it?

Scitator (changes requested):

catalyst/dl/experiments/experiment.py
hexfaker: `get_datasets_only` is implemented in `ConfigExperiment` and users should not reimplement it in most cases.  
May be we should update our config structure to separrate num workers, batch size etc from the rest of data params to make get_loaders and get_datasets independent?

catalyst/dl/scripts/trace.py
hexfaker: Getting loaders was in the first version. But it was tremendously slow, due to large batch and slow worker initialization. Another option is to override those params, but Im not sure whether  it's a better option.

Scitator: Even it is slow, I dont think it is an issue - you need to trace you model only once per several training usually. But some additional logic with `get_datasets_only` will force you to think about API each experiment :) (like "why do we need this method?")

another issue, with this new method - we will need to maintain it in future versions and check that it works correctly. 

solution with loaders does'nt have such issues (btw, you always can dynamically change `num_workers` and `batch_size` for speedup), that is why it looks more general for me

hexfaker: OK, this time you're right 😋

Scitator (approved):

PR #198, opened 2019-05-26
Multi-headed Value functions

KhrulkovV:
Implementation of the algorithm from [Hyperbolic Discounting and Learning over
Multiple Horizons](https://arxiv.org/pdf/1902.06865.pdf), as described in the section devoted to Atari experiments (using only the last head for playing, learning Q-functions for multiple gammas are interpreted as _auxiliary tasks_).


Scitator (changes requested):
based on slack discussion

catalyst/rl/agents/critic.py
```diff
@@ -43,6 +43,17 @@ def forward(self, state):
...
+    @property
+    def num_heads(self) -> int:
+        return self.head_net.num_heads
+
+    @property
+    def hyperbolic_constant(self) -> float:
+        if self.num_heads == 1:
+            return 1.0
+        else:
+            return self.head_net.hyperbolic_constant
```
Scitator: can you make it 1.0 for `num_heads == 1` case?

catalyst/rl/agents/critic.py
```diff
@@ -129,6 +140,17 @@ def distribution(self) -> str:
...
```
Scitator: maybe just always return self.head_net.hyperbolic_constant without any additional logic?

catalyst/rl/agents/head.py
```diff
@@ -13,14 +14,18 @@ def __init__(
         num_atoms: int = 1,
         bias: bool = False,
         distribution: str = None,
-        values_range: tuple = None
+        values_range: tuple = None,
+        num_heads=1,
+        hyperbolic_constant=0.01
```
Scitator: annotations, like float & int, please

catalyst/rl/agents/head.py
```diff
@@ -30,23 +35,43 @@ def __init__(
...
             # make critic outputs (B, 1) instead of (B, )
-            x = x.unsqueeze_(dim=1)
-        return x
+            x = [z.unsqueeze_(dim=1) for z in x]
+
+        # B x num_heads x num_outputs x num_atoms
+        # if num_heads == 1, squeeze it out not to crash other algorithms
+        return torch.stack(x, dim=1).squeeze(1)
+
+    def _build_head(
```
Scitator, line 51: place it just after the `__init__` please, to store all the private methods before the public one

catalyst/rl/offpolicy/algorithms/dqn.py
```diff
@@ -32,38 +41,65 @@ def _init(self):
             self.tau = self._to_tensor(tau)
             self._loss_fn = self._quantile_loss
 
+        # If we have only one head, unsqueeze its output for coherence
+        if self._num_heads == 1:
```
Scitator: during critic forward:
```
# B x num_heads x num_outputs x num_atoms
# if num_heads == 1, squeeze it out not to crash other algorithms
return torch.stack(x, dim=1).squeeze(1)
```
and now `unsqueeze(1)`?

what do you think about making critic always output `[bs; num_heads; num_outputs; num_atoms]` (so it would  always have output of deterministic shape)? and refactor all other algorithms to use correct shape (add/remove `squeeze/unsqueeze` blocks)?

catalyst/rl/offpolicy/algorithms/dqn.py
```diff
@@ -32,38 +41,65 @@ def _init(self):
...
+        rewards_t = rewards_t[:, None, :]
+        actions_t = actions_t.unsqueeze(1).repeat(1, self._num_heads, 1)
+        gammas = gammas[None, :, None]
+
+        q_values_t = critic(states_t).squeeze(-1).gather(-1, actions_t)
         q_values_tp1 = \
-            self.target_critic(states_tp1).squeeze(-1).max(-1, keepdim=True)[0]
-        q_target_t = rewards_t + (1 - done_t) * gamma_ * q_values_tp1.detach()
+            target_critic(states_tp1).squeeze(-1).max(-1, keepdim=True)[0]
+        q_target_t = rewards_t + (1 - done_t * gammas * q_values_tp1.detach())
```
Scitator: something strange with brackets happened here... can you please check the correctness?

catalyst/rl/offpolicy/algorithms/dqn.py
```diff
@@ -72,21 +108,30 @@ def _categorical_loss(
     def _quantile_loss(
         self, states_t, actions_t, rewards_t, states_tp1, done_t
     ):
-        gamma = self._gamma ** self._n_step
+
+        critic = self.critic_fn
```
Scitator: do not you mind to write shapes for all the tensors in `_base_loss/_categorical_loss/_quantile_loss` ?
because of many reshapes and transforms it's a little bit hard to understand what is going on without understanding tensor shapes

Scitator (changes requested):

catalyst/rl/agents/head.py
KhrulkovV: Fixed.

catalyst/rl/agents/critic.py
KhrulkovV: Fixed, thanks.

catalyst/rl/offpolicy/algorithms/dqn.py
KhrulkovV: Sounds good, will try to implement.

catalyst/rl/offpolicy/algorithms/dqn.py
KhrulkovV: Fixed, thanks.

Scitator (approved):

rfali:
@Scitator I can't seem to find files that were changed as part of this PR in this repo. I am not sure what I am missing. There is the older Catalyst Repo [v1](https://github.com/catalyst-team/catalyst-rl/blob/master/catalyst_rl/rl/offpolicy/algorithms/dqn.py) that has some parts of this. Where is the v2 dqn.py?

@KhrulkovV I have a question about this implementation. The way you have used the hyperbolic gammas on this [line](https://github.com/catalyst-team/catalyst-rl/blob/ec18ff4a58b6d00652f772231db8de86debb4b3d/catalyst_rl/rl/offpolicy/algorithms/dqn.py#L20) is different from the original [source code](https://github.com/google-research/google-research/tree/master/hyperbolic_discount) from the authors. Your implementation (not sure because of Catalyst RL v1 v2 confusion) uses these gammas for evaluating the Bellman updates, whereas these gammas are used by the authors to estimate the integral (see [here](https://github.com/google-research/google-research/blob/672c1857ba56c195000d0e0d38a46b10579b7c03/hyperbolic_discount/hyperbolic_dqn_agent.py#L119)). Since you did not implement the hyperbolic acting policy (integrate over all the heads) and used only the largest gamma (last head) for actions, while still computing loss over all heads, I think the gammas you obtain still need to raised to the k power as is done on this [line](https://github.com/google-research/google-research/blob/672c1857ba56c195000d0e0d38a46b10579b7c03/hyperbolic_discount/hyperbolic_dqn_agent.py#L130). These are the self.gammas that are used in the Bellman updates as [here](https://github.com/google-research/google-research/blob/672c1857ba56c195000d0e0d38a46b10579b7c03/hyperbolic_discount/hyperbolic_dqn_agent.py#L264). 

Just thought to run this by you, since I could definitely be wrong. I am implementing this paper in another library and this is the only other work that implements this technique. By the way, do you have an example of how I can use this PR to a simple gym environment?

rfali:
@KhrulkovV  please correct me if I am wrong, if I concatenate the q-values learned over all-heads (final layer), and then slice the resulting tensor to get individual q-values, and keep the rest as same (calculate loss for each and aggregate), that would still be equivalent?

PR #199, opened 2019-05-26
Segmented on policy

AlexGrinch:
Added functionality to collect segments of trajectories in on-policy methods instead of collecting complete episodes. To control the segment length, a parameter **segment_length** has been added to sampler. The default value is **None** which corresponds to segments of length **episode_len**.

Tested on LunarLander-v2 with **segment_length=128**.


Scitator (changes requested):
based on slack discussion

PR #214, opened 2019-06-05
Fix batch2device 

BloodAxe:
Currently, if dataset outputs a batch as follows:
`{ "features": (tensor1, tensor2), "targets": tensor3 }` 
A `tensor1` and `tensor2` will not be moved to GPU. 

New implementation ensures all nested tensors are properly moved to target device


catalyst/dl/experiments/core.py
```diff
@@ -114,11 +114,25 @@ def __init__(
         # additional
         self._check_run = False
 
+    def _any2device(self, value, device):
```
Scitator: what about move this function to UtilsFactory? it looks general enough and can be used not only in runner internals. 

btw, could you please update the code style to something like
```
return dict(
    (k, self._any2device(v, device)) 
    for k, v in value.items())
```
by this way it's much easier to read and understand

PR #223, opened 2019-06-11
python refactoring & project restructuration

Scitator:


catalyst/contrib/callbacks/confusion_matrix.py
```diff
@@ -0,0 +1,103 @@
+class ConfusionMatrixCallback(Callback):
```
TezRomacH: make sure you have updated docs for all methods that you moved

PR #238, opened 2019-06-19
MixupCallback

lightforever:
The callback for the mixup.

fields: features which must be affected by the mixup.

flpha: Beta distribution a=b parameters

train_only: boolean. Apply to train only? As the mixup use the proxy inputs, the targets are also proxy.
We are not interested in it, are we? So, if train_only is False, use a standard output/metric.

### DIFF
--- a/catalyst/dl/callbacks/mixup.py
+++ b/catalyst/dl/callbacks/mixup.py
@@ -1,0 +1,45 @@
+from typing import List
+from catalyst.dl.core.state import RunnerState
+from catalyst.dl.callbacks import CriterionCallback
+import numpy as np
+import torch
+
+
+class MixupCallback(CriterionCallback):
+    def __init__(self, fields: List[str] = ('features',), alpha=1.0, train_only=True, **kwargs):
+        assert alpha >= 0, 'alpha must be>=0'
+
+        super(MixupCallback, self).__init__(**kwargs)
+
+        self.train_only = train_only
+        self.fields = fields
+        self.alpha = alpha
+        self.lam = 1
+        self.index = None
+
+    def on_batch_start(self, state: RunnerState):
+        if self.train_only and state.loader_name != 'train':
+            return
+        if self.alpha > 0:
+            self.lam = np.random.beta(self.alpha, self.alpha)
+        else:
+            self.lam = 1
+
+        self.index = torch.randperm(state.input[self.fields[0]].shape[0])
+        if state.input[self.fields[0]].is_cuda:
+            self.index = self.index.cuda()
+
+        for f in self.fields:
+            state.input[f] = self.lam * state.input[f] + (1 - self.lam) * state.input[f][self.index]
+
+    def _compute_loss(self, state: RunnerState, criterion):
+        if self.train_only and state.loader_name != 'train':
+            return super(MixupCallback, self)._compute_loss(state, criterion)
+        pred = state.output[self.output_key]
+        y_a = state.input[self.input_key]
+        y_b = state.input[self.input_key][self.index]
+
+        return self.lam * criterion(pred, y_a) + (1 - self.lam) * criterion(pred, y_b)
+
+
+__all__ = ['MixupCallback']
--- a/catalyst/dl/callbacks/mixup.py
+++ b/catalyst/dl/callbacks/mixup.py
@@ -20,8 +20,14 @@ class MixupCallback(CriterionCallback):
 
         You may not use them together.
     """
-    def __init__(self, fields: List[str] = ("features",), alpha=1.0,
-                 train_only=True, **kwargs):
+
+    def __init__(
+            self,
+            fields: List[str] = ("features",),
+            alpha=1.0,
+            train_only=True,
+            **kwargs
+    ):
         """
         Args:
             fields (List[str]): list of features which must be affected.
--- a/catalyst/dl/callbacks/mixup.py
+++ b/catalyst/dl/callbacks/mixup.py
@@ -6,7 +6,10 @@ import torch
 
 
 class MixupCallback(CriterionCallback):
-    def __init__(self, fields: List[str] = ('features',), alpha=1.0, train_only=True, **kwargs):
+    def __init__(self, fields: List[str] = ('features',), alpha=1.0,
+                 train_only=True, **kwargs):
+        assert len(
+            fields) > 0, 'At least one field for MixupCallback is required'
         assert alpha >= 0, 'alpha must be>=0'
 
         super(MixupCallback, self).__init__(**kwargs)
@@ -30,7 +33,8 @@ class MixupCallback(CriterionCallback):
             self.index = self.index.cuda()
 
         for f in self.fields:
-            state.input[f] = self.lam * state.input[f] + (1 - self.lam) * state.input[f][self.index]
+            state.input[f] = self.lam * state.input[f] + (1 - self.lam) * \
+                             state.input[f][self.index]
 
     def _compute_loss(self, state: RunnerState, criterion):
         if self.train_only and state.loader_name != 'train':
@@ -39,7 +43,8 @@ class MixupCallback(CriterionCallback):
         y_a = state.input[self.input_key]
         y_b = state.input[self.input_key][self.index]
 
-        return self.lam * criterion(pred, y_a) + (1 - self.lam) * criterion(pred, y_b)
+        return self.lam * criterion(pred, y_a) + (1 - self.lam) * criterion(
+            pred, y_b)
 
 
 __all__ = ['MixupCallback']
--- a/catalyst/dl/callbacks/mixup.py
+++ b/catalyst/dl/callbacks/mixup.py
@@ -9,8 +9,31 @@ from catalyst.dl.callbacks import CriterionCallback
 
 
 class MixupCallback(CriterionCallback):
+    """
+    Callback to do mixup augmentation.
+
+    Paper: https://arxiv.org/abs/1710.09412
+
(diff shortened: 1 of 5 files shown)


catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,45 @@
...
+from catalyst.dl.callbacks import CriterionCallback
+import numpy as np
+import torch
+
+
+class MixupCallback(CriterionCallback):
+    def __init__(self, fields: List[str] = ('features',), alpha=1.0, train_only=True, **kwargs):
+        assert alpha >= 0, 'alpha must be>=0'
+
+        super(MixupCallback, self).__init__(**kwargs)
```
Scitator: for py3 you can use just `super().__init__` to follow lib code standarts

catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,45 @@
...
+
+        self.train_only = train_only
+        self.fields = fields
+        self.alpha = alpha
+        self.lam = 1
+        self.index = None
+
+    def on_batch_start(self, state: RunnerState):
+        if self.train_only and state.loader_name != 'train':
```
Scitator: it better to use `state.loader_name.startswith('train')`

PS. btw, please use " to follow lib code standarts

catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,45 @@
...
+        self.index = torch.randperm(state.input[self.fields[0]].shape[0])
+        if state.input[self.fields[0]].is_cuda:
+            self.index = self.index.cuda()
+
+        for f in self.fields:
+            state.input[f] = self.lam * state.input[f] + (1 - self.lam) * state.input[f][self.index]
+
+    def _compute_loss(self, state: RunnerState, criterion):
+        if self.train_only and state.loader_name != 'train':
+            return super(MixupCallback, self)._compute_loss(state, criterion)
```
Scitator: can be simplified to `super().` (py3)

catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,45 @@
...
+    def on_batch_start(self, state: RunnerState):
+        if self.train_only and state.loader_name != 'train':
+            return
+        if self.alpha > 0:
+            self.lam = np.random.beta(self.alpha, self.alpha)
+        else:
+            self.lam = 1
+
+        self.index = torch.randperm(state.input[self.fields[0]].shape[0])
+        if state.input[self.fields[0]].is_cuda:
```
Scitator: what do you think about such simplification:`self.index = self.index.to(state.device)`?

catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,45 @@
...
+            state.input[f] = self.lam * state.input[f] + (1 - self.lam) * state.input[f][self.index]
+
+    def _compute_loss(self, state: RunnerState, criterion):
+        if self.train_only and state.loader_name != 'train':
+            return super(MixupCallback, self)._compute_loss(state, criterion)
+        pred = state.output[self.output_key]
+        y_a = state.input[self.input_key]
+        y_b = state.input[self.input_key][self.index]
+
+        return self.lam * criterion(pred, y_a) + (1 - self.lam) * criterion(pred, y_b)
```
Scitator: to allow easy debug for future needs it's better to make something like
```
loss = ...
return loss
```

catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,45 @@
+from typing import List
+from catalyst.dl.core.state import RunnerState
+from catalyst.dl.callbacks import CriterionCallback
+import numpy as np
+import torch
+
+
+class MixupCallback(CriterionCallback):
+    def __init__(self, fields: List[str] = ('features',), alpha=1.0, train_only=True, **kwargs):
```
Scitator: please, update the docs for new callback

catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,45 @@
+from typing import List
+from catalyst.dl.core.state import RunnerState
+from catalyst.dl.callbacks import CriterionCallback
+import numpy as np
+import torch
```
Scitator, line 5: to follow lib code standarts, its better to use the following import sequence:
- core python libs (typing, collection, etc)
- python libs (numpy, etc)
- dl libs (pytorch, etc)
- catalyst imports

Scitator (changes requested):

catalyst/dl/callbacks/mixup.py
lightforever: That is great information, thanks. Corrected

catalyst/dl/callbacks/mixup.py
lightforever: Fixed. What is the reason to use double quotes instead of a single one? 
To type " we should use Shift on the most keyboards. Do you have one that allows doing the type in a single press ?)

catalyst/dl/callbacks/mixup.py
lightforever: That is great information, thanks. Corrected

catalyst/dl/callbacks/mixup.py
lightforever: This is much better. Corrected

catalyst/dl/callbacks/mixup.py
lightforever: Fixed

catalyst/dl/callbacks/mixup.py
lightforever: Done

catalyst/dl/callbacks/mixup.py
lightforever: Done, thanks.

catalyst/dl/callbacks/mixup.py
```diff
@@ -1,50 +1,56 @@
...
+        super().__init__(**kwargs)
 
         self.train_only = train_only
         self.fields = fields
         self.alpha = alpha
         self.lam = 1
         self.index = None
 
+    def _is_needed(self, state: RunnerState):
+        return not self.train_only or state.loader_name.startswith("train")
```
Scitator: what about check this flag `on_loader_start`?

catalyst/dl/callbacks/mixup.py
```diff
@@ -1,50 +1,56 @@
...
+
     def on_batch_start(self, state: RunnerState):
-        if self.train_only and state.loader_name != 'train':
+        if not self._is_needed(state):
```
Scitator: what about check this flag `on_loader_start`?

catalyst/dl/callbacks/mixup.py
```diff
@@ -1,50 +1,56 @@
...
+
 import torch
 
+from catalyst.dl.core.state import RunnerState
+from catalyst.dl.callbacks import CriterionCallback
+
 
 class MixupCallback(CriterionCallback):
-    def __init__(self, fields: List[str] = ('features',), alpha=1.0,
+    def __init__(self, fields: List[str] = ("features",), alpha=1.0,
```
Scitator: add docs in [Google format](https://github.com/catalyst-team/catalyst/blob/master/CONTRIBUTING.md#documentation) please

catalyst/dl/callbacks/mixup.py
lightforever: Thanks, done.

catalyst/dl/callbacks/mixup.py
lightforever: Thanks, done.

catalyst/dl/callbacks/mixup.py
lightforever: Thanks, done.

catalyst/dl/callbacks/mixup.py
```diff
@@ -0,0 +1,81 @@
...
+    Paper: https://arxiv.org/abs/1710.09412
+
+    Note:
+        MixupCallback is inherited from CriterionCallback and
+        does its work.
+
+        You may not use them together.
+    """
+    def __init__(self, fields: List[str] = ("features",), alpha=1.0,
+                 train_only=True, **kwargs):
```
Scitator: could you please refactor to 
```
def __init__(
    self, 
    fields: List[str] = ("features",), 
    alpha=1.0,
    train_only=True, 
    **kwargs
):
```
one more common lib standard :)

lightforever: Done :)

Scitator:
It looks great!
As a final change, could you please add config3.yml - a copy of config1.yml with `MixupCallback` (rather than `CriterionCallback`) to travis CI?

All you need is to add
```
- PYTHONPATH=./examples:./catalyst:${PYTHONPATH}
          python catalyst/dl/scripts/run.py
          --expdir=./examples/_tests_mnist_stages2
          --config=./examples/_tests_mnist_stages2/config3.yml
          --logdir=./examples/logs/_tests_mnist_stages2
          --check
```
after https://github.com/catalyst-team/catalyst/blob/master/.travis.yml#L94

and contribute config3.yml with your `MixupCallback`

PS. config1 could be found [here](https://github.com/catalyst-team/catalyst/blob/master/examples/_tests_mnist_stages2/config1.yml)

lightforever:
@Scitator, Done :)

Scitator (approved):

PR #239, opened 2019-06-19
`FORCE_JPEG_TURBO`

bagxi:


catalyst/utils/image.py
```diff
@@ -2,6 +2,7 @@
 import os
 import numpy as np
 import imageio
+import jpeg4py as jpeg
```
hexfaker: Import should be moved to lines 35-36 (inside condition) because now it's optional and it's absence in the environment should not cause any errors

requirements.txt
```diff
@@ -3,6 +3,7 @@ numpy>=1.16.0
 torch>=0.4.1
 torchvision>=0.2.1
 imageio
+jpeg4py
```
hexfaker: This is optional dependency and should not be present in requirements

hexfaker (changes requested):

hexfaker (approved):

Scitator:
try-except is quite slow in python and can level all jpeg4py speed.
So, my suggestion is to follow [this ods thread](https://opendatascience.slack.com/archives/C044B7CSQ/p1560240300231300)
For example, you can use something like [this](https://github.com/catalyst-team/catalyst/blob/master/catalyst/utils/compression.py#L9) to handle import only once and store some global constant like `JPEG4PY_ENABLED`. Nevertheless, do not remember to check jpeg4py import and decode methods before enabling this flag :)

Scitator (changes requested):
we need something like `JPEG4PY_ENABLED` to speedup this solution

catalyst/utils/image.py
```diff
@@ -9,6 +11,31 @@
 _IMAGENET_STD = (0.229, 0.224, 0.225)
 _IMAGENET_MEAN = (0.485, 0.456, 0.406)
 
+logger = logging.getLogger(__name__)
+
+JPEG4PY_ENABLED = False
+if os.environ.get('FORCE_JPEG_TURBO', False):
```
Scitator: please use " to follow lib standarts

catalyst/utils/image.py
```diff
@@ -9,6 +11,31 @@
...
+        JPEG4PY_ENABLED = True
+    except ImportError:
+        logger.warning(
+            "jpeg4py not available. "
+            "To install jpeg4py, run `pip install jpeg4py`."
+        )
+    except OSError:
+        logger.warning(
+            "jpeg4py not available. "
+            "To install libjpeg-turbo, see `http://bfy.tw/OC7m`."
```
Scitator: I think, it's better to link jpeg4py and libjpeg-turbo installation directly

catalyst/utils/image.py
```diff
@@ -29,7 +56,10 @@ def imread(uri, grayscale=False, expand_dims=True, rootpath=None):
             uri if uri.startswith(rootpath) else os.path.join(rootpath, uri)
         )
 
-    img = imageio.imread(uri, as_gray=grayscale, pilmode="RGB")
+    if JPEG4PY_ENABLED and rootpath.endswith(("jpg", "JPG", "jpeg", "JPEG")):
```
Scitator: `rootpath`? or `uri`?

Scitator (changes requested):

catalyst/utils/image.py
```diff
@@ -29,7 +56,10 @@ def imread(uri, grayscale=False, expand_dims=True, rootpath=None):
             uri if uri.startswith(rootpath) else os.path.join(rootpath, uri)
         )
 
-    img = imageio.imread(uri, as_gray=grayscale, pilmode="RGB")
+    if JPEG4PY_ENABLED and uri.endswith(("jpg", "JPG", "jpeg", "JPEG")):
+        img = jpeg.JPEG(uri).decode()
```
Scitator, line 60: you also need to force grayscale if specified

Scitator (changes requested):

PR #247, opened 2019-06-30
RawObservationWrapper

KhrulkovV:
Added wrapper to track the original transitions before the preprocessing step


catalyst/rl/core/sampler.py
```diff
@@ -219,6 +225,12 @@ def _run_sample_loop(self):
 
             if not self._infer or self._force_store:
                 self._store_trajectory(trajectory)
+                if "raw_trajectory" in trajectory_info:
+                    self._store_trajectory(
```
Scitator: what about:
```python
raw_trajectory = trajectory_info.pop("raw_trajectory")
self._store_trajectory(raw_trajectory, raw=True)
```
?

catalyst/rl/core/trajectory_sampler.py
```diff
@@ -164,5 +164,10 @@ def sample(self, exploration_strategy=None):
 
         trajectory = self.get_trajectory()
         trajectory_info = {"reward": reward, "num_steps": num_steps}
+        if info and "raw_trajectory" in info:
+            trajectory_info["raw_trajectory"] = info["raw_trajectory"]
+            reward = np.sum(info["raw_trajectory"][2])
+        trajectory_info["raw_reward"] = reward
+
         assert all(len(x) == num_steps for x in trajectory)
```
Scitator, line 177: maybe, we should add len assert for `raw_trajectory` too?

catalyst/rl/db/mongo.py
```diff
@@ -42,13 +43,16 @@ def get_sample_flag(self) -> bool:
...
-    def push_trajectory(self, trajectory):
+    def push_trajectory(self, trajectory, raw=False):
         trajectory = utils.structed2dict_trajectory(trajectory)
         trajectory = utils.pack(trajectory)
-        self._trajectory_collection.insert_one({
-            "trajectory": trajectory,
-            "date": datetime.datetime.utcnow(),
-            "epoch": self._epoch
+        collection = self._raw_trajectory_collection if raw else \
+            self._trajectory_collection
```
Scitator: can we edit codestyle a bit?
```python
collection = self._raw_trajectory_collection if raw 
    else self._trajectory_collection
```

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,83 @@ def __getitem__(self, i):
...
 
+class RawObservationWrapper(Wrapper):
+    def __init__(
+            self,
+            env,
+            allow_early_resets=False,
+            reset_keywords=(),
+            info_keywords=()
+    ):
+        Wrapper.__init__(self, env=env)
```
Scitator: maybe you mean
```python
super().__init__(env)
```
?

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,83 @@ def __getitem__(self, i):
...
+                raise ValueError(
+                    "Expected you to pass kwarg %s into reset" % k
+                )
+            self.current_reset_info[k] = v
+        initial_state = self.env.reset(**kwargs)
+        self.observations = [initial_state, ]
+        self.actions = []
+        self.rewards = []
+        self.dones = []
+        return self.env.reset(**kwargs)
```
Scitator:
```python
return initial_state
```
?

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,83 @@ def __getitem__(self, i):
         return self._force()[..., i]
 
 
+class RawObservationWrapper(Wrapper):
```
Scitator: maybe `RawTrajectoryWrapper`?

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,83 @@ def __getitem__(self, i):
...
+    def update(self, ob, rew, done, action, info):
+        self.observations.append(ob)
+        self.actions.append(action)
+        self.rewards.append(rew)
+        self.dones.append(done)
+        assert isinstance(info, dict)
+        if done:
+            self.needs_reset = True
+            info["raw_trajectory"] = (
+                np.array(self.observations)[:-1],
```
Scitator: have you tested that with Dict observations?

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,83 @@ def __getitem__(self, i):
...
+
+    def update(self, ob, rew, done, action, info):
+        self.observations.append(ob)
+        self.actions.append(action)
+        self.rewards.append(rew)
+        self.dones.append(done)
+        assert isinstance(info, dict)
+        if done:
+            self.needs_reset = True
+            info["raw_trajectory"] = (
```
Scitator: btw, I think, it would be easier to understand if you would return `info`.
inner update could cause a lot of misunderstanding

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,83 @@ def __getitem__(self, i):
...
+                np.array(self.observations)[:-1],
+                np.array(self.actions),
+                np.array(self.rewards),
+                np.array(self.dones)
+            )
+
+    def get_total_steps(self):
```
Scitator: deprecated?

Scitator (changes requested):

catalyst/rl/core/trajectory_sampler.py
KhrulkovV: fixed

catalyst/rl/db/mongo.py
KhrulkovV: fixed

examples/atari/src/atari_wrappers.py
KhrulkovV: fixed

examples/atari/src/atari_wrappers.py
KhrulkovV: fixed

examples/atari/src/atari_wrappers.py
KhrulkovV: fixed

examples/atari/src/atari_wrappers.py
KhrulkovV: I have tested it with the Minigrid env, seems to be working

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,72 @@ def __getitem__(self, i):
...
 
 
+class RawTrajectoryWrapper(Wrapper):
+    def __init__(
+            self,
+            env,
+            allow_early_resets=False,
+            reset_keywords=(),
+            info_keywords=()
+    ):
```
Scitator: any docs, please?

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,72 @@ def __getitem__(self, i):
...
+
+    def reset(self, **kwargs):
+        self.reset_state()
+        for k in self.reset_keywords:
+            v = kwargs.get(k)
+            if v is None:
+                raise ValueError(
+                    "Expected you to pass kwarg %s into reset" % k
+                )
+            self.current_reset_info[k] = v
```
Scitator: why do we need this?

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,72 @@ def __getitem__(self, i):
...
+        self.actions = []
+        self.rewards = []
+        self.dones = []
+        return initial_state
+
+    def reset_state(self):
+        if not self.allow_early_resets and not self.needs_reset:
+            raise RuntimeError(
+                "Wrap your env with "
+                "RawObservationWrapper(env, allow_early_resets=True)"
```
Scitator: `RawTrajectoryWrapper` already :)

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +305,72 @@ def __getitem__(self, i):
...
+        self.reset_state()
+        for k in self.reset_keywords:
+            v = kwargs.get(k)
+            if v is None:
+                raise ValueError(
+                    "Expected you to pass kwarg %s into reset" % k
+                )
+            self.current_reset_info[k] = v
+        initial_state = self.env.reset(**kwargs)
+        self.observations = [initial_state, ]
```
Scitator: what do you think about optimizing the code with DynamicArray?
just like it was done on [`trajectory_sampler`](https://github.com/catalyst-team/catalyst/blob/master/catalyst/rl/core/trajectory_sampler.py#L37)
this should unite buffers logic and make it easy to support it in future

examples/atari/src/atari_wrappers.py
```diff
@@ -351,7 +418,7 @@ def make_atari_env(
     # env = MaxAndSkipEnv(env, skip=4)
     if max_episode_steps is not None:
         env = TimeLimit(env, max_episode_steps=max_episode_steps)
-
+    env = RawObservationWrapper(env)
```
Scitator: `RawTrajectoryWrapper`?

Scitator (changes requested):

catalyst/rl/core/trajectory_sampler.py
```diff
@@ -162,13 +162,17 @@ def sample(self, exploration_strategy=None):
         if not self._sample_flag.value:
             return None, None
 
+        print(num_steps)
```
Scitator: debug print?

examples/atari/src/atari_wrappers.py
Scitator: for the RawTrajectoryWrapper arguments :)

Scitator: something like, what do we need `reset_keywords ` and `info_keywords ` for?

catalyst/rl/core/sampler.py
```diff
@@ -224,8 +230,14 @@ def _run_sample_loop(self):
             if trajectory is None:
                 continue
 
+            raw = "raw_trajectory" in trajectory_info
+            # Do it firsthand, so the loggers don't crush
+            if raw:
+                raw_trajectory = trajectory_info.pop("raw_trajectory")
```
Scitator: what about 
```
raw_trajectory = trajectory_info.pop("raw_trajectory", None)
```
and then check `if raw_trajectory is not None`?

examples/atari/src/atari_wrappers.py
```diff
@@ -304,6 +307,124 @@ def __getitem__(self, i):
         return self._force()[..., i]
 
 
+class RawTrajectoryWrapper(Wrapper):
```
Scitator: can you please add this wrapper to `catalyst.rl.wrappers`?
you need to create new module dir with `__init__.py` and `raw_trajectory.py`

I think this wrapper would be used very much.

catalyst/rl/wrappers/__init__.py
```diff
@@ -0,0 +1,11 @@
+# flake8: noqa
+
+from .agent import ActorSpec, CriticSpec
```
Scitator: todo: fix copy-paste :)

KhrulkovV: oops

catalyst/rl/wrappers/raw_trajectory.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class RawTrajectoryWrapper(Wrapper):
+    def __init__(
+            self,
+            env,
+            allow_early_resets=False,
+            initial_capacity=1e3,
+            reset_keywords=(),
+            info_keywords=()
```
Scitator: why do we need `reset_keywords` and `info_keywords`?

KhrulkovV: `reset_keywords` are needed if it is required to pass some arguments to the reset function, like `env.reset(difficulty=1)`, so it checks if all the required arguments are passed. And `info_keywords` seems not to be required.

Scitator: why should this wrapper check reset arguments?
I think, this feature should be somewhere in the EnvWrapper [here](https://github.com/catalyst-team/catalyst/blob/master/catalyst/rl/environment/gym.py#L55), for example

PR #248, opened 2019-07-01
Print LR metric in scientific notation

BloodAxe:
Fixes `_base/lr=0.0000` in logs when LR is 1e-5 and lower


catalyst/dl/utils/formatters.py
```diff
@@ -44,12 +44,21 @@ class TxtMetricsFormatter(MetricsFormatter):
     def __init__(self):
         super().__init__("[{asctime}] ")
 
+    def _format_metric(self, name, value):
+        if name in {"_base/lr"}:
```
Scitator: what if we would print in scientific notation not only LR, but all the metrics that are below 1e-4?

PR #266, opened 2019-07-24
Feature/checkpoint metrics

TezRomacH:
Added `_metrics.json` for all saved checkpoints in logdir

Снимок экрана 2019-07-30 в 14 39 54

for every single `.pth` show all `valid_metrics` 
Снимок экрана 2019-07-30 в 14 38 37

except for `last` checkpoint, all others are sorted by `main_metric`


catalyst/dl/callbacks/checkpoint.py
```diff
@@ -77,8 +84,12 @@ def save_checkpoint(
             last_filepath = last_item[0]
             os.remove(last_filepath)
 
-    def pack_checkpoint(self, **kwargs):
-        return utils.pack_checkpoint(**kwargs)
+        checkpoints = [
+            (Path(filepath).stem, metric)
+            for (filepath, _, metric) in self.top_best_metrics
```
Scitator: `metric` should be `valid_metrics`, correctness remark

catalyst/dl/callbacks/checkpoint.py
```diff
@@ -77,8 +84,12 @@ def save_checkpoint(
...
+        ]
+        metrics = OrderedDict(checkpoints + [("last", valid_metrics)])
```
Scitator: maybe you should also save "best" checkpoint metrics? so it would be the same as checkpoints naming

catalyst/dl/callbacks/checkpoint.py
Scitator: but do not confuse it with current epoch `valid_metrics`

catalyst/dl/callbacks/checkpoint.py
```diff
@@ -96,13 +107,15 @@ def on_epoch_end(self, state: RunnerState):
...
+        valid_metrics = dict(state.metrics.valid_values)
+
+        checkpoint = utils.pack_checkpoint(
             model=state.model,
             criterion=state.criterion,
             optimizer=state.optimizer,
             scheduler=state.scheduler,
             epoch_metrics=dict(state.metrics.epoch_values),
-            valid_metrics=dict(state.metrics.valid_values),
+            valid_metrics=valid_metrics,
```
Scitator, line 124: if you change `valid_metrics`... maybe it's better to do the same `for epoch_metrics`?

TezRomacH: why? i use `valid_metrics` for my calculations, and don't use `epoch_metrics`

Scitator: you don't use `valid_metrics` in this scope

TezRomacH: updated

TezRomacH:
updated the code style

PR #268, opened 2019-07-25
RL/mongo with safitty

TezRomacH:
`safitty` allows you to remove a lot of `None` checks

Снимок экрана 2019-07-25 в 12 43 39


catalyst/rl/db/mongo.py
```diff
@@ -3,6 +3,8 @@
 from catalyst.rl import utils
 from catalyst.rl.core import DBSpec
 
+import safitty
```
Scitator: all libraries imports should be on top of internal catalyst ones

TezRomacH: fixed!

PR #272, opened 2019-07-28
fix compartability with projector

vaklyuenkov:
right input data format to add embeddings for projector


catalyst/contrib/scripts/project_embeddings.py
```diff
@@ -110,7 +110,7 @@ def main(args, _=None):
     summary_writer = SummaryWriter(args.out_dir)
     summary_writer.add_embedding(
         features,
-        metadata=df.values,
+        metadata=df[meta_header].astype(str).values,
```
Scitator, line 113: we need to check that meta_header is not none

Scitator (changes requested):

Scitator:
@vaklyuenkov Please, fix the codestyle for merging.

PR #274, opened 2019-07-30
Feature/auto accuracy callback

TezRomacH:


catalyst/dl/callbacks/metrics/accuracy.py
```diff
@@ -4,6 +4,25 @@
 from catalyst.dl.utils import criterion
 
 
+def _get_list_args(num_classes: int) -> List[int]:
```
Scitator: maybe, `_get_default_accuracy_args`?

TezRomacH: no, because it used also in `MapKCallback`

Scitator: mAP is accuracy still

Scitator: we just need better naming, rather than such abstract one

TezRomacH: don't want to name it `accuracy`, because if we add some new classification callback this func will be useful

Scitator: you place it in the `metrics/accuracy.py` and make it protected, so it looks  okay to use it only for the accuracy

TezRomacH: ok

PR #278, opened 2019-07-31
predict_loader feature

Scitator:


catalyst/dl/experiment/base.py
```diff
@@ -35,7 +35,15 @@ def __init__(
     ):
         self._model = model
         self._loaders = loaders
-        self._callbacks = callbacks or []
+
+        if callbacks is None:
+            callbacks = OrderedDict()
+        elif isinstance(callbacks, List):
+            callbacks = OrderedDict([
+                (i, value)
```
TezRomacH, line 43: in the typing, you said `OrderedDict[str, Callback]`, but you set integer keys.
```suggestion
                (str(i), value)
```

Scitator: it's backward compatibility with List indexing

TezRomacH: so  change `Union[OrderedDict[str, Callback], List[Callback]]` to

`Union[OrderedDict[Union[str, int], Callback], List[Callback]]`

PR #281, opened 2019-08-02
Feature on_exception for Callbacks

TezRomacH:


catalyst/dl/callbacks/logging.py
```diff
@@ -163,4 +173,17 @@ def on_loader_end(self, state: RunnerState):
             )
 
 
-__all__ = ["VerboseLogger", "ConsoleLogger", "TensorboardLogger"]
+class RaiseExceptionCallback(Callback):
```
Scitator: because it is in `logging.py`, maybe we should rename it to `ExceptionLogger`?

TezRomacH: It's complicated because it doesn't log anything. It is just re-raise any exception if it needed

PR #292, opened 2019-08-09
Added param `state_dict` param for the resnet-encoder

TezRomacH:
Can be very useful for segmentation:
- if you want to use pretrain different than `ImageNet`
- on Kaggle with internet-off kernels

Example:
```python
from catalyst.contrib.models.segmentation import ResnetUnet

encoder_params = dict(state_dict="/path/to/model/resnet18-5c106cde.pth")
model = ResnetUnet(num_classes=1, pretrained=False, encoder_params=encoder_params)
```


TezRomacH:
I have tested it on my computer :)

catalyst/contrib/models/segmentation/encoder/resnet.py
```diff
@@ -37,12 +37,33 @@ def __init__(
         arch: str = "resnet18",
         pretrained: bool = True,
         requires_grad: bool = None,
-        layers_indices: List[int] = None
+        layers_indices: List[int] = None,
+        state_dict: dict = None,
```
Scitator: what do you think about `Path` support for this arg?
It would be easier to use from Config API.

TezRomacH: ok :)

Scitator:
Btw, would you like to also add such functionality to [ResnetEncoder](https://github.com/catalyst-team/catalyst/blob/master/catalyst/contrib/models/encoder/resnet.py#L8)?

TezRomacH:
done

PR #293, opened 2019-08-09
Remove bias weight decay feature

TezRomacH:


catalyst/dl/utils/torch.py
```diff
@@ -96,7 +96,48 @@ def get_loader(
...
+        >>> model = ResnetUnet()
+        >>> params = process_model_params(model, weight_decay=0.2)
+        >>> optimizer = torch.optim.Adam(params, lr=0.0003)
+    """
+    params = list(model.named_parameters())
+
+    if not remove_bias_decay or (weight_decay == 0.0):
+        return [param for (name, param) in params]
+
+    # no bias decay from https://arxiv.org/pdf/1812.01187.pdf
```
Scitator: abs please

catalyst/dl/experiment/config.py
```diff
@@ -162,10 +163,14 @@ def _get_optimizer(self, *, model_params, **params):
         return optimizer
 
     def get_optimizer(self, stage: str, model: nn.Module) -> _Optimizer:
-        model_params = model.parameters()
         optimizer_params = \
             self.stages_config[stage].get("optimizer_params", {})
 
+        weight_decay: float = optimizer_params.get("weight_decay", 0.0)
+        remove_bias_decay = optimizer_params.pop("remove_bias_decay", True)
```
Scitator: maybe `no_bias_weight_decay`, it's better to have the name as simple as possible

catalyst/dl/utils/torch.py
```diff
@@ -96,7 +96,48 @@ def get_loader(
...
+        iterable: parameters for an optimizer
+
+    Examples:
+        >>> model = ResnetUnet()
+        >>> params = process_model_params(model, weight_decay=0.2)
+        >>> optimizer = torch.optim.Adam(params, lr=0.0003)
+    """
+    params = list(model.named_parameters())
+
+    if not remove_bias_decay or (weight_decay == 0.0):
```
Scitator: for float values we should use `np.isclose`

catalyst/dl/utils/torch.py
```diff
@@ -96,7 +96,48 @@ def get_loader(
     return loader
 
 
+def process_model_params(
```
Scitator: what do you think about `group_model_params` name?
or `gather_*` / `arrange_*`

btw, catalyst naming `process_something` suggests inplace change of the args. kwargs, rather then returning some result (for methods with result, it's better to use `get_*`)

TezRomacH: i'll name it `get_*` then

PR #297, opened 2019-08-13
rename get_model_params and update README

TezRomacH:
- Added to README our tutorials and pipelines
- fixed `_get_default_accuracy_args`. Previously if we only had 5 classes we saw constant `accuracy05 = 100%` which is not necessary.
- renamed `get_model_params` to `process_model_params` to comply with the Catalyst's standards.
- added `pretrained: str = "imagenet"` to the tutorial
- added `verbose=True` flag to `predict_loader`
- added predicted class name after `predict_loader`
- fixed `plotly` in google colab

### DIFF
--- a/examples/notebooks/classification-tutorial.ipynb
+++ b/examples/notebooks/classification-tutorial.ipynb
@@ -19,7 +19,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "scrolled": true
+   },
    "outputs": [],
    "source": [
     "!pip install -U catalyst\n",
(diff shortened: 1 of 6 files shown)


examples/notebooks/classification-tutorial.ipynb
```diff
@@ -1306,25 +1306,34 @@
    ]
   }
  ],
-  "metadata": {
-    "kernelspec": {
-      "display_name": "Python 3",
-      "language": "python",
-      "name": "python3"
+ "metadata": {
```
Scitator, line 1309: do not change meta please

PR #319, opened 2019-08-21
add checkpoint name to trace

TezRomacH:
add checkpoint name to `catalyst-dl trace`
```
catalyst-dl trace cifar_simple -c stage1.0
```

or 
```
catalyst-dl trace cifar_simple --checkpoint=last --out-dir /path/to/output/
```

----

Tracing in `train` mode
```
catalyst-dl trace cifar_simple --checkpoint best --mode train
```

Tracing with grads
```
catalyst-dl trace cifar_simple --with-grad
```

### DIFF
--- a/catalyst/dl/scripts/trace.py
+++ b/catalyst/dl/scripts/trace.py
@@ -64,13 +64,20 @@ def build_args(parser: ArgumentParser):
     parser.add_argument(
         "--checkpoint", "-c",
         default="best",
-        help="Checkpoint's name to trace"
+        help="Checkpoint's name to trace",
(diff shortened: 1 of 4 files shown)


Scitator:
What do you think about an option of model tracing in train state / required-grad state?

TezRomacH:
great idea, i'll add this!

Scitator:
Could you please add an option to set model name by user?
something like
```bash
catalyst-dl trace cifar_simple --checkpoint best --mode train --with-grad --out-model=./my_traced_model.pth
```
Looks like it gives more flexibility for users

TezRomacH:
ok

PR #321, opened 2019-08-22
Dl/extra scripts

Scitator:


catalyst/contrib/scripts/resize_images.py
```diff
@@ -0,0 +1,68 @@
...
+        self.target = target
+        self.source = source
+        self.interpolation = interpolation
+        self.max_side = max_side
+
+    def preprocess(self, image_path: Path):
+        target_path = self.target / image_path.relative_to(self.source)
+        target_path.parent.mkdir(parents=True, exist_ok=True)
+
+        image = np.array(imageio.imread(image_path))
```
TezRomacH: can be changed to `catalyst.utils.imread`

catalyst/data/__main__.py
```diff
@@ -61,13 +61,14 @@
...
 from catalyst.__version__ import __version__
-from catalyst.contrib.scripts import check_images, \
+from catalyst.contrib.scripts import check_images, resize_images, \
     image2embedding, tag2label, split_dataframe
 
 COMMANDS = OrderedDict(
     [
         ("tag2label", tag2label),
         ("check-images", check_images),
+        ("resize-images", resize_images),
```
TezRomacH: what do you think about adding `clear_exif` also?

TezRomacH: as `("clear-exif", clear_exif),`

PR #335, opened 2019-08-28
Feature/paths dataset

TezRomacH:


Scitator:
Could you please add some tests for this case?

PR #337, opened 2019-08-28
External monitoring tools

Scitator:
Demo: https://app.wandb.ai/scitator/classification

### DIFF
--- a/catalyst/dl/__init__.py
+++ b/catalyst/dl/__init__.py
@@ -9,4 +9,4 @@ from .core import Experiment, Runner, RunnerState, \
 from .experiment import BaseExperiment, SupervisedExperiment, ConfigExperiment
 from .meters import AverageValueMeter, ClassErrorMeter, ConfusionMeter, \
     MSEMeter, MovingAverageValueMeter, AUCMeter, APMeter, mAPMeter
-from .runner import SupervisedRunner
+from .runner import *
--- a/catalyst/dl/runner/__init__.py
+++ b/catalyst/dl/runner/__init__.py
@@ -6,7 +6,7 @@ from .supervised import SupervisedRunner
 
 try:
     import wandb
-    from .wandb import WandbRunner
+    from .wandb import WandbRunner, SupervisedWandbRunner
 except ImportError:
     logger.warning(
         "wandb not available, switching to pickle. "
--- a/catalyst/dl/runner/__init__.py
+++ b/catalyst/dl/runner/__init__.py
@@ -1,3 +1,4 @@
 # flake8: noqa
 
 from .supervised import SupervisedRunner
+from .wandb import *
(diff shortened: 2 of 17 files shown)


catalyst/dl/__init__.py
```diff
@@ -9,4 +9,4 @@
 from .experiment import BaseExperiment, SupervisedExperiment, ConfigExperiment
 from .meters import AverageValueMeter, ClassErrorMeter, ConfusionMeter, \
     MSEMeter, MovingAverageValueMeter, AUCMeter, APMeter, mAPMeter
-from .runner import SupervisedRunner
+from .runner import *
```
TezRomacH, line 12: we have to decide our code-style about `*`.

previously, we make an agreement to not use asterisks

catalyst/dl/runner/__init__.py
```diff
@@ -1,3 +1,4 @@
 # flake8: noqa
 
 from .supervised import SupervisedRunner
+from .wandb import *
```
TezRomacH: also

TezRomacH: maybe it's better to move this
```
try:
    import wandb
    WANDB_ENABLED = True
except ImportError:
    logger.warning(
        "wandb not available, switching to pickle. "
        "To install wandb, run `pip install wandb`."
    )
    WANDB_ENABLED = False

if WANDB_ENABLED:
    from .wandb import WandbRunner
```
here to simplify `catalyst/dl/runner/wandb.py` code?

catalyst/dl/runner/wandb.py
```diff
@@ -0,0 +1,87 @@
...
+except ImportError:
+    logger.warning(
+        "wandb not available, switching to pickle. "
+        "To install wandb, run `pip install wandb`."
+    )
+    WANDB_ENABLED = False
+
+
+if WANDB_ENABLED:
+    class WandbRunner(Runner):
```
TezRomacH: please, add some documentations

catalyst/utils/misc.py
```diff
@@ -76,6 +76,17 @@ def append_dict(dict1, dict2):
     return dict1
 
 
+def flatten_dict(d, parent_key="", sep="/"):
```
TezRomacH, line 79: is it better fit in Safitty?

TezRomacH (changes requested):

catalyst/dl/__init__.py
Scitator: It would be nice if you could propose workaround. For now - it looks like bad, but working solution.

TezRomacH: Personally, I prefer to write `__all__` in the script and then in init actually use `from .something import *`

TezRomacH: But i wonder why you demanded not to write `*` 🤔 in my previous PRs

PR #343, opened 2019-08-30
PoC: callback ordering

TezRomacH:


catalyst/dl/core/callback.py
```diff
@@ -7,12 +7,12 @@
...
-    Scheduler = 200
-    Metric = 300
-    Logger = 400
-    Other = 2000
+    Criterion = 20
+    Optimizer = 40
+    Scheduler = 60
+    Metric = 80
+    Logger = 100
+    Other = 101
```
TezRomacH: I think it has to be bigger because we may want to order our loggers

PR #361, opened 2019-09-10
Feature. Criterion Aggregator

TezRomacH:
In some cases, it is necessary to pass a dictionary to the loss, not a value by a specific key. 
For example, in detection, the loss may be counted with several keys.


catalyst/dl/callbacks/optimizer.py
```diff
@@ -73,26 +81,38 @@ def on_epoch_start(self, state):
...
         if isinstance(loss, list):
-            loss = torch.mean(torch.stack(loss))
+            if len(loss) > 1:
+                logger.warning(
+                    f"Attention, the loss is a list. "
+                    f"Only the last value will be used for `backward`."
+                    f"To aggregate losses into "
+                    "one value use `CriterionAggregatorCallback`"
+                )
+            loss = loss[-1]
```
Scitator: why do we take last loss by default?

catalyst/dl/callbacks/criterion.py
```diff
@@ -1,44 +1,74 @@
...
+        return result
 
     def _compute_loss(self, state: RunnerState, criterion):
-        loss = criterion(
-            state.output[self.output_key], state.input[self.input_key]
-        )
+        output = self._get(state.output, self.output_key)
+        input = self._get(state.input, self.input_key)
+
+        loss = criterion(output, input)
```
Scitator, line 78: as I remember you required to add KV-support. From my perspective, something like:
```python
def _get(dictionary: dict, key: Optional[str]) -> Any:
    result = {key_:dictionary[key_] for key_ in key} if key is not None else dictionary
    return result

output = self._get(state.output, self.output_key)
input = self._get(state.input, self.input_key)

loss = criterion(**output, **input)
```
should work

catalyst/dl/callbacks/criterion.py
```diff
@@ -57,7 +87,79 @@ def on_batch_end(self, state: RunnerState):
...
+class CriterionAggregatorCallback(Callback):
+    """
+    This callback allows you to aggregate the values of the loss
+    (by ``sum`` or ``mean``) and put the value back into ``state.loss``.
+    """
+    def __init__(
+        self,
+        prefix: str,
+        loss_keys: List[str] = None,
+        loss_aggregate_fn: str = "sum",
```
Scitator, line 109: by default, it's better to use mean aggregation to prevent batch size dependency

catalyst/dl/callbacks/optimizer.py
TezRomacH: Because in case of a list, the last loss is appended to the and
```
        if state.loss is not None:
            if isinstance(state.loss, list):
                state.loss.append(loss)
            else:
                state.loss = [state.loss, loss]
```

catalyst/dl/callbacks/criterion.py
TezRomacH: u mean, add support for multiple `self.output_key` and `self.input_key`?

catalyst/dl/callbacks/criterion.py
TezRomacH: disagree, the sum of the losses is more frequent (e.g. adding different regularizations)

TezRomacH: if the size is a reason, you could add a multiplier to each loss and the aggregated loss

catalyst/dl/callbacks/criterion.py
Scitator: yes, as u say "in detection, the loss may be counted with several keys."

catalyst/dl/callbacks/optimizer.py
Scitator: and why only last one? I think it is much better to sum/mean them, or just throw an exception, because last-loss usage looks very tricky

TezRomacH: fixed

catalyst/dl/callbacks/criterion.py
TezRomacH: done !

TezRomacH:
fixed

catalyst/dl/callbacks/criterion.py
```diff
@@ -57,7 +92,79 @@ def on_batch_end(self, state: RunnerState):
...
+
+class CriterionAggregatorCallback(Callback):
+    """
+    This callback allows you to aggregate the values of the loss
+    (by ``sum`` or ``mean``) and put the value back into ``state.loss``.
+    """
+    def __init__(
+        self,
+        prefix: str,
+        loss_keys: List[str] = None,
```
Scitator: maybe, 
```python
Union[str, List[str]]
```

catalyst/dl/callbacks/criterion.py
```diff
@@ -57,7 +92,79 @@ def on_batch_end(self, state: RunnerState):
...
+            loss_aggregate_fn (str): function for aggregation.
+                Must be either ``sum`` or ``mean``.
+            multiplier (float): scale factor for the aggregated loss.
+        """
+        super().__init__(CallbackOrder.Criterion + 1)
+        assert prefix is not None and isinstance(prefix, str), \
+            "prefix must be str"
+        self.prefix = prefix
+
+        if not isinstance(loss_keys, list) and loss_keys is not None:
```
Scitator: maybe,
```python
if isinstance(loss_keys, str):
```

catalyst/dl/callbacks/criterion.py
```diff
@@ -57,7 +92,79 @@ def on_batch_end(self, state: RunnerState):
...
+        if loss_aggregate_fn == "sum":
+            self.loss_fn = lambda x: torch.sum(torch.stack(x)) * multiplier
+        elif loss_aggregate_fn == "mean":
+            self.loss_fn = lambda x: torch.mean(torch.stack(x)) * multiplier
+        else:
+            raise ValueError("loss_aggregate_fn must be `sum` or `mean`")
+
+        self.loss_aggregate_name = loss_aggregate_fn
+
+    def _get_losses(self, loss: Any) -> List[torch.Tensor]:
```
Scitator: maybe,
```python
_process_loss
_preprocess_loss
_aggregate_loss
```

just because you already have loss in params, so.... there should not be any `get` :)

catalyst/dl/callbacks/optimizer.py
```diff
@@ -73,26 +81,36 @@ def on_epoch_start(self, state):
...
+        if isinstance(loss, dict):
+            if len(loss) > 1:
+                error = "Loss is a dict, to aggregate losses into " \
+                        "one value use `CriterionAggregatorCallback`."
+                if self.loss_key is None:
+                    error = error + " Or try to pass `loss_key` " \
+                                    "in the OptimizerCallback init"
+                raise ValueError(error)
+            loss = list(loss.values())[-1]
+        return loss
```
Scitator, line 102: why do you throw an exception for List case and take last loss for dict?
I suggest to remove this method and use only
```python
loss = state.get_key(key="loss", inner_key=self.loss_key)
```
all aggregation logic should be done in another callback

PR #365, opened 2019-09-12
Multiple models & optimizers & phases support. Vanilla GAN example on MNIST.

reyzharkov:
Finally! GANs supported! 
For now as a standalone example, which may be generalized in the future to any multi-model & multi-optimizer & multi-phase experiments


catalyst/dl/callbacks/phase.py
```diff
@@ -0,0 +1,112 @@
+from typing import List
+from collections import OrderedDict
+
+from catalyst.dl.core import Callback, CallbackOrder
+
+
+class Phase:
```
TezRomacH, line 7: docstring please

catalyst/dl/callbacks/phase.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self, name: str = None, steps: int = None):
+        self.steps = int(steps) if steps is not None else None
+        self.curr_step = 0
+        self.name = name
+
+
+class PhaseManager:
```
TezRomacH, line 20: docstring

catalyst/dl/core/state.py
```diff
@@ -87,8 +90,9 @@ def __init__(
         self.timer = TimerManager()
 
         # base metrics
-        self.lr = None
-        self.momentum = None
+        single_optimizer = isinstance(optimizer, Optimizer)
+        self.lr = None if single_optimizer else defaultdict(lambda: None)
```
TezRomacH, line 94: dont use `defaultdict`, try https://github.com/catalyst-team/safitty from catalyst-ecosystem

examples/mnist_gans/runner.py
```diff
@@ -0,0 +1,68 @@
+from typing import Mapping, Any
+
+import torch
+from catalyst.dl import Runner
+
+
+class GANRunner(Runner):
```
TezRomacH: maybe it's better to move `GANRunner` from examples into `catalyst/dl/runner` ?

reyzharkov: OK

Actually right now to implement some other GAN basing on current vanilla GANRunner someone will have to override it almost completely. I suppose we will have to think about better/more generic/easy extendable implementation which will be more useful parent class.

But I'm going to left this important task to other PR's =)

catalyst/dl/core/state.py
reyzharkov: So far I believe defaultdicts are much more clear here with much less code.
I agree that dicts instead may prevent some errors (but without safitty, again)

Maybe you can give me a masterclass?
```python
        self.lr = None
        self.momentum = None
        if isinstance(optimizer, dict):
            self.lr = dict()
            self.momentum = dict()
            for key in optimizer:
                safitty.set(self.lr, key, value=None)
                safitty.set(self.momentum, key, value=None)
```

catalyst/dl/callbacks/phase.py
reyzharkov: done

catalyst/dl/callbacks/phase.py
reyzharkov: done. BTW initially I wanted to reply "no documentation is kind of catalyst culture=(", but when I merged, I've seen that things are changing slowly, that's great!

catalyst/dl/callbacks/phase.py
```diff
@@ -0,0 +1,126 @@
...
+        return self.valid_phases[self.valid_index].name
+
+
+class PhaseManagerCallback(Callback):
+    """
+    PhaseManagerCallback updates state.phase
+    """
+
+    VM_ALL = "all"  # (in validation) use all callbacks
+    VM_SAME = "same"  # (in validation) same phases as in training
```
Scitator: let's use full naming, like `VALIDATION_MODE_ALL`
btw, it's better to use Enum for such cases

catalyst/dl/experiment/config.py
```diff
@@ -111,14 +111,28 @@ def _preprocess_model_for_stage(self, stage: str, model: _Model):
...
+        key_value_flag = params.pop("_key_value", False)
+
+        if key_value_flag:
+            model = {}
+            for key, params_ in params.items():
+                model[key] = ConfigExperiment._get_model(**params_)
+        else:
+            model = MODELS.get_from_params(**params)
+            if model is not None and torch.cuda.is_available():
+                model = model.cuda()
```
Scitator: you don't need to do this here, 
1. `.cuda` is a bit deprecated
2. Runner has a logic with mode.to(device)

catalyst/dl/experiment/config.py
```diff
@@ -139,15 +153,74 @@ def get_criterion(self, stage: str) -> _Criterion:
...
+        optimizer_key = params.pop("optimizer_key", None)
         optimizer = OPTIMIZERS.get_from_params(**params, params=model_params)
 
         if load_from_previous_stage:
             checkpoint_path = f"{self.logdir}/checkpoints/best_full.pth"
             checkpoint = utils.load_checkpoint(checkpoint_path)
-            utils.unpack_checkpoint(checkpoint, optimizer=optimizer)
+
+            dict2load = optimizer
+            if optimizer_key:
```
Scitator: use `is not None` please, it's python

catalyst/dl/runner/__init__.py
```diff
@@ -5,6 +5,7 @@
 logger = logging.getLogger(__name__)
 
 from .supervised import SupervisedRunner
+from .gan_runner import GANRunner
```
Scitator: we can simply name it `gan`

catalyst/dl/experiment/config.py
```diff
@@ -290,14 +348,23 @@ def get_loaders(self, stage: str) -> "OrderedDict[str, DataLoader]":
...
         return loaders
 
+    @staticmethod
+    def _get_callback(**params):
+        wrapper_params = params.pop("_wrapper", None)
+        callback = CALLBACKS.get_from_params(**params)
+        if wrapper_params:
+            wrapper_params["base_callback"] = callback
+            return ConfigExperiment._get_callback(**wrapper_params)
+        return callback
```
Scitator, line 356: what do you think about another approach, like here https://github.com/catalyst-team/catalyst/blob/master/catalyst/contrib/optimizers/lookahead.py#L84 ?
so in the config it should looks like:
```
callback_name:
    callback: WrapperCallback
    param_1: ...
    param_2: ...
    base_callback_params:
        param_3: ...
        param_4: ...

```

catalyst/dl/runner/gan_runner.py
```diff
@@ -0,0 +1,71 @@
...
+import torch
+from catalyst.dl import Runner
+
+
+class GANRunner(Runner):
+    def __init__(
+        self,
+        model=None,
+        device=None,
+        images_key="images",
```
Scitator: to make it more general, let's use `features` as a key?

catalyst/dl/core/runner.py
```diff
@@ -166,11 +169,17 @@ def _run_epoch(self, loaders):
                 self._run_loader(loader)
             self._run_event("loader_end")
 
+    def _run_prestage(self, stage: str):
```
Scitator, line 172: looks like a dirty hack :)

examples/mnist_gans/callbacks.py
```diff
@@ -0,0 +1,124 @@
...
+        self._n_row = n_row
+        self._reset()
+
+    def _reset(self):
+        self._loader_batch_count = 0
+        self._loader_visualized_in_current_epoch = False
+
+    @staticmethod
+    def _get_tensorboard_logger(state: RunnerState) -> SummaryWriter:
+        for logger_name, logger in state.loggers.items():
```
Scitator: you can simply import `state.loggers["tensorboard"]`

catalyst/dl/experiment/config.py
reyzharkov: I think my solution needs 1 line in config.yml compared to 2 lines here

catalyst/dl/core/runner.py
reyzharkov: I created this method to memorize something in runner before stage begins but after initialization is done. It's obviously a little bit hacky, but I think it's still useful, i.e. the hack is not dirty=)

examples/mnist_gans/callbacks.py
reyzharkov: and what will happen when you rename that logger? =(
what do you think about the new solution? it seems to become even less readable

catalyst/dl/runner/gan_runner.py
reyzharkov: ok

catalyst/dl/runner/__init__.py
reyzharkov: ok, done

catalyst/dl/experiment/config.py
reyzharkov: done

catalyst/dl/experiment/config.py
reyzharkov: I remember it was in the previous version of _get_model, so I left it just in case
Anyway, I removed it

catalyst/dl/callbacks/phase.py
reyzharkov: ok

Scitator:
somehow moved to https://github.com/catalyst-team/catalyst/pull/407
:)

PR #372, opened 2019-09-18
Feature/catalyst dl init

TezRomacH:
To initialize empty project run
```bash
catalyst-dl init
```

to initialize with a pipeline use
```bash
catalyst-dl init --pipeline classification
```

or 

```bash
catalyst-dl init -p segmentation
```

Thanks for the idea, @hexfaker 👍 

minuses: this function starts to argue on the wandb
Снимок экрана 2019-09-25 в 20 56 03


hexfaker:
You are welcome)

catalyst/dl/scripts/init.py
```diff
@@ -0,0 +1,83 @@
+import argparse
+import shutil
+from pathlib import Path
+from git import Repo
+
+
+def _copy(input_dir: Path, output_dir: Path):
```
Scitator: looks like some general one –> utils?

examples/_empty/src/model.py
```diff
@@ -0,0 +1,12 @@
+import torch
+from catalyst.contrib import registry
+
+
+@registry.Model
+class Model(torch.nn.Module):
```
Scitator, line 6: it would be nice if this example would work just out of the box
You already have datasets, so I suggest you to add CIFAR10 model here

examples/_empty/src/experiment.py
```diff
@@ -0,0 +1,43 @@
...
+        result = transforms.Compose(
+            [
+                transforms.ToTensor(),
+                transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))
+            ]
+        )
+
+        return result
+
+    def get_datasets(self, stage: str, **kwargs):
```
Scitator: it would be nice to add here some arguments to give users an intuition about `data_params` from config

examples/_empty/src/model.py
Scitator: btw, even better idea – to add `get_from_params` method to have an example of model construction from config

Scitator (approved):

Issue #366, opened 2019-09-13
How to resume training from checkpoint?


bamps53:
I found resume argument in predict_loader function, but not in train function. How can I resume from best checkpoint??

ngxbac:
Very good question!. 
I think Catalyst hasnt supported it, is it? @Scitator 
I encountered this problem before. Here is my solution.

1. In the configure file (ex: config.yml), you should to add `resume` to the config:
```yml
args:
  expdir: "src"
  logdir: &logdir "logs"
  baselogdir: "logs"
  resume: #your checkpoint
``` 
2. Modify your `runner.py` 
```python

class ModelRunner(Runner):
    def predict_batch(self, batch: Mapping[str, Any]):
        output = self.model(batch["images"])
        return {
            "logits": output
        }

    def _run_stage(self, stage: str):
        self._prepare_state(stage)
        loaders = self.experiment.get_loaders(stage)
        self.callbacks = self.experiment.get_callbacks(stage)

        self._run_event("stage_start")
        for epoch in range(18, self.state.num_epochs):
            self.state.stage_epoch = epoch

            self._run_event("epoch_start")
            self._run_epoch(loaders)
            self._run_event("epoch_end")

            if self._check_run and self.state.epoch >= 3:
                break
            if self.state.early_stop:
                self.state.early_stop = False
                break

            self.state.epoch += 1
        self._run_event("stage_end")
```

In here, I resume from `epoch 18`.

Scitator:
Hi,

Sorry for long answer – there are few conferences and competitions now :)

So, it's really good question... or even proposal.
For now, you can use workaround like (Notebook API)
```
from catalyst.dl import SupervisedRunner
from catalyst import utils

model = Net()
checkpoint = utils.load_checkpoint(resume=...)
utils.unpack_checkpoint(checkpoint, model=model)

runner = SupervisedRunner(model=model)
runner.train(...)
```

For config API, all you need to do is something like
```
callbacks_params:
      loader:
        callback: CheckpointCallback
        resume: <path/to/checkpoint.pth>
```

Scitator:
merged, should work now :)

bamps53:
Thanks, it's working correctly now :)

Scitator:
Easy, you always can ping us

- [Telegram](https://t.me/catalyst_team)
- [Gitter](https://gitter.im/catalyst-team/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
[...]

PR #377, opened 2019-09-24
Supervised runner resume feature

Scitator:


Scitator:
issue: https://github.com/catalyst-team/catalyst/issues/366

catalyst/dl/runner/supervised.py
```diff
@@ -162,6 +164,17 @@ def train(
         if model is not None:
             self.model = model
 
+        if resume is not None:
+            callbacks = callbacks or OrderedDict()
+            checkpoint_callback_flag = all([
```
TezRomacH: are you sure about `all` instead of `any`?

Scitator: you are right

PR #381, opened 2019-09-26
fix - AccuracyCallback for multi-label classification

lightforever:
criterion/accuracy now can work only with a multi-class classification problem.

Use case: multi-label classification problem

targets: batch_size x num_classes, each sample has binary vector where more than one 1 is possible

outputs: batch_size x num_classes, logits

criterion: BCEWithLogitsLoss

To cope with that, we can check targets.size(1), if it's > 1 that is a multi-class classification problem.

topk can not be applied in that case.

### DIFF
--- a/catalyst/dl/callbacks/metrics/accuracy.py
+++ b/catalyst/dl/callbacks/metrics/accuracy.py
@@ -37,27 +37,19 @@ class AccuracyCallback(MultiMetricCallback):
     Accuracy metric callback.
 
     It can be used either for
-        - multi-class task, you can use accuracy_args. threshold and activation
-        are not required.
-            - input_key point on tensor: batch_size
-            - output_key point on tensor: batch_size x num_classes
-
-            An example of a configuration(config API):
-                accuracy:
-                  callback: AccuracyCallback
+        - multi-class task:
+            -you can use accuracy_args.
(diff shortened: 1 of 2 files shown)


Scitator:
Could you please describe the usecase for such feature?
And why dont use just sever AccuracyCallbacks?
like [here](https://github.com/catalyst-team/classification/blob/master/configs/exp_splits_augs.yml#L118)

or... Are you talking about multi-label classification problem?

lightforever:
Pardon, yes I am talking about multi-label classification problem.

Use case: 

targets: batch_size x num_classes, each sample has binary vector where more than one 1 is possible

outputs:  batch_size x num_classes, logits

criterion: BCEWithLogitsLoss

Callback configuration:
```
    accuracy:
      callback: AccuracyCallback
      threshold: 0.5
      activation: Sigmoid
```

Scitator:
Hmm, now I see, looks reasonable.
Nevertheless, as @TezRomacH says, could you please write mini docs for `accuracy`, because now it looks a bit complicated. For example, there is no way to use `topk` in you case and it can be confusing.

lightforever:
@Scitator , I have added docs to Accuracy callback and criterion/accuracy.

How do you think, is it enough?

PR #389, opened 2019-09-28
remove duplicating code

TezRomacH:


catalyst/utils/time.py
```diff
@@ -0,0 +1,20 @@
+from datetime import datetime
+
+
+def get_time_string(format: str = None) -> str:
```
Scitator: maybe, `get_utcnow_time`? `get_time_string` looks too abstract

PR #399, opened 2019-10-02
Fix an issue with setting seed for different TF versions

yulits:
Fixed an error with setting seed for TF2.* and added a condition to avoid warning for TF1.14, 1.15


valyukov:
Sure. Check this.

PR #411, opened 2019-10-06
Fix Epoch Number for Console Logging

ostamand:
Fix epoch number in console logging. 

Before: 

Screen Shot 2019-10-06 at 1 17 12 PM

After: 

Screen Shot 2019-10-06 at 1 14 38 PM

### DIFF
--- a/bin/check_dl.sh
+++ b/bin/check_dl.sh
@@ -24,8 +24,8 @@ fi
 python -c """
 from safitty import Safict
 metrics=Safict.load('$LOGFILE')
-assert metrics.get('stage1.2', 'loss') < metrics.get('stage1.0', 'loss')
-assert metrics.get('stage1.2', 'loss') < 2.0
+assert metrics.get('stage1.3', 'loss') < metrics.get('stage1.1', 'loss')
+assert metrics.get('stage1.3', 'loss') < 2.0
 """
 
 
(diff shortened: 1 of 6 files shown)


Scitator:
Hi, thanks for PR!

Nevertheless, working with checkpoint indices, we also need to logs correctness. In your example, best model saved as `train.1.pth`, that means checkpoint at 1 epoch.... but from logs perspective it's already 2nd epoch – a little bit confusing :)

Scitator (changes requested):
+

ostamand:
You are right. I missed that. Since all the logs are epoch zero based, I propose that we keep it as is but update the total number of epochs in the console progress bar. To keep it consistent.

Screen Shot 2019-10-07 at 9 04 40 AM

Let me know if you would prefer a full refactoring of the epoch number so that it starts at 1 everywhere & I will do that instead.

Scitator:
Conceptually, I think it's much better to make full refactoring and save model after N epoch as `checkpoint.N.pth`. It's much easier to understand :)

Looks like you need to refactor [here](github.com/catalyst-team/catalyst/blob/master/catalyst/dl/core/runner.py#L197L198) and [here](https://github.com/catalyst-team/catalyst/blob/master/catalyst/dl/core/runner.py#L141)... at least I hope so :) 
We still need to check logs, metrics and overall correctness.

ostamand:
@Scitator let me know what you think of this: 

Added `epoch_log` & `stage_epoch_log` properties to `RunnerState` to keep it cleaner.

***Multi-Stage Training Example***

_Stage 1:_

image

[...]

image

_Stage 2:_

image

_Loading from Stage 2 Checkpoint:_

image

***Looking at the plots***

_Stage 1:_

image

_Stage 2:_

image

_Loading from Stage 2 Checkpoint:_

image

ostamand:
Was not expecting travis to fail for such a small change. Let me look into it first. Sorry.

Scitator:
Hah, @ostamand looks like you also need to rewrite tests a bit :)
We have check like `train.1.loss` < `train.0.loss`.... but with your new indexing... we need +1 there

ostamand:
@Scitator Cool thanks for the tests commit! I was about to take a look at it. Let me know if there is any other change you want me to do?

Issue #445, opened 2019-10-13
Feature request: log more system metrics with W&B


ternaus:
It would be nice to add logging for 

`learning rate`, `momentum`, `batch_time`, `data_time` to Weights & Biases

lightforever:
This issue will be closed after https://github.com/catalyst-team/catalyst/pull/458

PR #458, opened 2019-10-18
adding learning rate, momentum, batch_time, data_time to Weights & Biases

lightforever:
## Description

adding learning rate, momentum, batch_time, data_time to Weights & Bi

## Related Issue

https://github.com/catalyst-team/catalyst/issues/445

## Type of Change

service metrics which are started with underline symbol ```_```  were renamed to their analogues without underline symbol

wandb used first symbol ```_``` for it service purposes, because of that fact, we can not send original metric names

### DIFF
--- a/catalyst/dl/runner/wandb.py
+++ b/catalyst/dl/runner/wandb.py
@@ -17,7 +17,8 @@ class WandbRunner(Runner):
     @staticmethod
     def _log_metrics(metrics: Dict, mode: str, suffix: str = ""):
         def key_locate(key: str):
-            """wandb uses first symbol _ for it service purposes
+            """
+            wandb uses first symbol _ for it service purposes
             because of that fact, we can not send original metric names
 
             Args:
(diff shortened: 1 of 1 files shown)


catalyst/dl/runner/wandb.py
```diff
@@ -16,8 +16,19 @@ class WandbRunner(Runner):
     """
     @staticmethod
     def _log_metrics(metrics: Dict, mode: str, suffix: str = ""):
+        def key_locate(key: str):
+            """
+            wandb used first symbol _ for it service purposes,
+            because of that fact, we can not send original metric names
+            :param key: metric name
```
TezRomacH: Please convert the documentation to Google Style :)

TezRomacH (changes requested):

catalyst/dl/runner/wandb.py
lightforever: Done)

TezRomacH (approved):

PR #496, opened 2019-11-05
Chore/verbose errors

gazay:
## Description

> This is discussion Pull Request, it will help to decide whether Catalyst needs this kind of feature and if yes - what could be standardized and which values could use defaults and which should throw an error.

There are lots of problems with invalid configs for Catalyst, and it's not clear why something is happening without looking into sources.
Right now, you can smoothly "run" experiment, and it will show you progress and stuff but will do nothing or fail after the first validation stage.

Here are a few examples:

The smallest "working" config:

```yaml
model_params:
  model: MNISTNet

args:
  expdir: src
  logdir: logs

stages:
  stage1:
    empty: 1
```

It will do nothing at all, but no exception will be shown.

Let's add `num_workers: 0` (there is no default value, that's why it's not even started previously):

```yaml
model_params:
  model: MNISTNet

args:
  expdir: src
  logdir: logs

stages:
  data_params:
    num_workers: 0
  stage1:
    empty: 1
```

It will show progress of `1/1` epoch and will fail after validation with assertion error:

```bash
$ catalyst-dl run --config=configs/smallest.yml --verbose
1/1 * Epoch (train): 100% 100/100 [00:00<00:00, 830.36it/s]
1/1 * Epoch (valid): 100% 100/100 [00:00<00:00, 1316.28it/s]
Traceback (most recent call last):
  File "/home/gazay/anaconda3/envs/py37/bin/catalyst-dl", line 11, in <module>
    load_entry_point('catalyst', 'console_scripts', 'catalyst-dl')()
  File "/home/gazay/code/catalyst/catalyst/dl/__main__.py", line 44, in main
    COMMANDS[args.command].main(args, uargs)
  File "/home/gazay/code/catalyst/catalyst/dl/scripts/run.py", line 85, in main
    runner.run_experiment(experiment, check=args.check)
  File "/home/gazay/code/catalyst/catalyst/dl/core/runner.py", line 263, in run_experiment
    self._run_event("exception", moment=None)
  File "/home/gazay/code/catalyst/catalyst/dl/core/runner.py", line 116, in _run_event
    getattr(logger, fn_name)(self.state)
  File "/home/gazay/code/catalyst/catalyst/dl/callbacks/misc.py", line 155, in on_exception
    raise exception
  File "/home/gazay/code/catalyst/catalyst/dl/core/runner.py", line 260, in run_experiment
    self._run_stage(stage)
  File "/home/gazay/code/catalyst/catalyst/dl/core/runner.py", line 234, in _run_stage
[...]

PR #505, opened 2019-11-06
Chore/default callbacks for ConfigExperiment

gazay:
## Description

I've added few tests on experiments and few default callbacks to `ConfigExperiment`.

I think that some of the callbacks should be set by default in non-infer mode in `SupervisedExperiment` and `ConfigExperiment`.

Right now, you can skip setting of `Criterion` and `Optimizer` for `Supervised` and `Config` experiments. Still, I think it could be right only for very rare edge cases in non-infer mode (I can't think of one).

In `SupervisedExperiment`, there is a check that if `Criterion` set in `non-infer` mode - then we should set default `CriterionCallback` if there is none. I think this kind of flexibility is not suitable for `Supervised` and `Config` experiments as absence of those callbacks makes no sense the same as absence of `Criterion` and `Optimizer` itself - without them, the experiment will be just iteration through data.

As result:

1. It should be almost impossible to run training without optimization process (if user forget to set criterion for example or misspelled key for it in config file) from `Supervised` and from `Config` experiments without exception being raised.
2. Users will set `callback_params` in config file only when they need to change something from defaults.

## Related Issue

https://github.com/catalyst-team/catalyst/pull/496

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] I have checked the code-style using `make check-style`.

### DIFF
--- a/catalyst/dl/experiment/config.py
+++ b/catalyst/dl/experiment/config.py
@@ -13,7 +13,8 @@ from torch.utils.data import (  # noqa F401
 from catalyst.dl import utils
 from catalyst.dl.callbacks import (
     CheckpointCallback, ConsoleLogger, CriterionCallback, OptimizerCallback,
-    RaiseExceptionCallback, SchedulerCallback, TensorboardLogger, VerboseLogger
+    RaiseExceptionCallback, SchedulerCallback, TensorboardLogger,
+    VerboseLogger
 )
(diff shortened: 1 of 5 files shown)


Scitator:
Merged, everything looks great, thanks for PR!

PR #535, opened 2019-11-28
added weighted sum

TezRomacH:
## Description

- Weighted sum fo `CriterionAggregatorCallback`
- Update tutorial

## Related Issue

N/A

## Type of Change

## Checklist

### DIFF
--- a/catalyst/utils/dataset.py
+++ b/catalyst/utils/dataset.py
@@ -53,9 +53,8 @@ def create_dataset(
     return dataset
 
 
-def split_dataset(
-    dataset: pd.DataFrame, **train_test_split_args
-) -> Tuple[DictDataset, DictDataset]:
+def split_dataset(dataset: pd.DataFrame,
+                  **train_test_split_args) -> Tuple[DictDataset, DictDataset]:
     """
     Split dataset in train and test parts.
 
@@ -167,7 +166,9 @@ def split_dataframe(
 
 
 def default_fold_split(
-    dataframe: pd.DataFrame, random_state: int = 42, n_folds: int = 5
+    dataframe: pd.DataFrame,
+    random_state: int = 42,
+    n_folds: int = 5
 ) -> pd.DataFrame:
     """
     Splits DataFrame into `N` folds.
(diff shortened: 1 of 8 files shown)


BloodAxe:
I think this can be already achieved by using existing `multiplier` property of `CriterionCallback`. One can set arbitrary multipliers to each criterion callback and achieve same effect of "mean", "sum" or "weighted sum" aggregation. And this would not add unnecessary complexity to `CriterionAggregationCallback`.

TezRomacH:
Yes, this is the right remark, the same functionality can be implemented via scale.

But, as we have noticed in ODS, in this case the metric in the terminal and the tensorboard will also scales. And this is not good, for example, if you multiply by a small factor like 0.0001

catalyst/utils/dataset.py
```diff
@@ -53,9 +53,8 @@ def create_dataset(
     return dataset
 
 
-def split_dataset(
-    dataset: pd.DataFrame, **train_test_split_args
-) -> Tuple[DictDataset, DictDataset]:
+def split_dataset(dataset: pd.DataFrame,
```
TezRomacH: weird, but it's from flake8

catalyst/utils/dataset.py
```diff
@@ -316,9 +317,8 @@ def balance_classes(
     return balanced_df
 
 
-def prepare_dataset_labeling(
-    dataframe: pd.DataFrame, class_column: str
-) -> Dict[str, int]:
+def prepare_dataset_labeling(dataframe: pd.DataFrame,
```
TezRomacH: also from flake8

.travis.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     - stage: Tests
       install:
         - *requirements_all
-        - pip install flake8 flake8-quotes yapf isort
+        - pip install -U flake8 flake8-quotes yapf isort
```
Scitator: what do you think about dumping the test packages versions for both local and CI tests?

catalyst/utils/dataset.py
Scitator: looks like flake8 version change, we need to fix this... issue :)

Scitator:
The PR looks good, I will try to dump the tests dependencies for CI correctness.

Scitator:
@TezRomacH  could you please merge with master? we have fixed the CI tests issue

catalyst/utils/dataset.py
TezRomacH: done

TezRomacH:
updated

lightforever:
@TezRomacH , please ensure you have the libraries in the list: https://github.com/catalyst-team/catalyst/blob/master/requirements/requirements-dev.txt

The versions must be the same. Especially, yapf must have version 0.28.0. The build works with this configuration.

PR #550, opened 2019-12-06
Added samplers_params

TezRomacH:
## Description

- Added config key for `samplers_params`

## Related Issue

N/A

## Type of Change

## Checklist

### DIFF
--- a/catalyst/dl/experiment/config.py
+++ b/catalyst/dl/experiment/config.py
@@ -18,7 +18,7 @@ from catalyst.dl.callbacks import (
 )
 from catalyst.dl.core import Callback, Experiment
 from catalyst.dl.registry import (
-    CALLBACKS, CRITERIONS, MODELS, OPTIMIZERS, SCHEDULERS
+    CALLBACKS, CRITERIONS, MODELS, OPTIMIZERS, SAMPLERS, SCHEDULERS
 )
 from catalyst.utils.typing import Criterion, Model, Optimizer, Scheduler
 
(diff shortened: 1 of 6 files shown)


TezRomacH:
@Scitator @bagxi could you please add your review?

gazay (approved):
I don't see any problems with this code but without tests it's hard to be sure. Btw it would be great to add these sampler params to wizard as well, but I can do it in separate PR

lightforever (approved):
Everythings seems reasonable

catalyst/dl/experiment/config.py
```diff
@@ -345,6 +351,17 @@ def get_loaders(self, stage: str) -> "OrderedDict[str, DataLoader]":
             assert isinstance(overridden_loader_params, dict), \
                 f"{overridden_loader_params} should be Dict"
 
+            sampler_params = samplers_params.pop(name, None)
```
Scitator, line 354: we should also check, that user's datasets and samplers have same names,

for example, what would be if we have
datasets: `{"train": ..., "valid": ...}` and samplers `{"train": ..., "valid000": ...}`?

Scitator: btw, the same logic should be implemented with loaders

TezRomacH: No, you don't. It's already done by default. 

You can see above that we are only iterated by the names of the datasets `for name, ds_ in datasets.items()`. And for each of them, we're trying to get a loader and a sampler

Scitator: please, add a validation check for our Catalyst.todo for this case

PR #585, opened 2020-01-09
Proof Of Concept for GAN in Notebook API

Arquestro:
## Description

This PoC is one of the first steps in Catalyst.GAN roadmap. This PR is supposed to be discussion\breaking, but for now it's only adding new features for Catalyst typing and GAN handling. MNIST GAN config example is used as a guide example for constructing GanExperiment and updating GanRunner.

## Type of Change

## Checklist


Arquestro:
It's in a state of WIP for now. 
Several TODOs:
- [x] Gan Experiment 
- [x] Gan Runner train
- [ ] Experiment tests(TODO Later, next PR)
- [x] Doc-strings for every new class and method

catalyst/utils/typing.py
```diff
@@ -1,12 +1,23 @@
...
 
 import torch
 from torch import nn, optim
+from torch.utils import data
 
-Model = nn.Module
+Model = Union[nn.Module, Dict[str, nn.Module]]
 Criterion = nn.Module
-Optimizer = optim.Optimizer
+Optimizer = Union[optim.Adam, optim.SGD]
```
TezRomacH: but, why?

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,122 @@
...
+    Optimizer,
+    Scheduler,
+    Dataset
+)
+
+
+class GanExperiment(Experiment):
+    def __init__(
+        self,
+        models: Union[Model, Dict[str, Model]],
```
TezRomacH: in 
```
def get_model(self, stage: str) -> Model:
        return self._models[stage]
```
you suppose that `models` can be only a dict

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,122 @@
...
+        datasets: Dict[str, Any] = {},
+        loaders: Dict[str, Any] = {},
+        state_params: Dict[str, Dict[str, Any]] = {},
+        monitoring_params: Dict = {},
+        distributed_params: Dict = {},
+        stages: Iterable[str] = ["train"],
+        logdir: str = "./logs",
+        verbose: bool = False,
+        initial_seed: int = 42,
+    ):
```
TezRomacH, line 22: also, we need docstrings for the CI

catalyst/dl/runner/gan.py
```diff
@@ -1,21 +1,59 @@
...
+)
 
 
 class GanRunner(Runner):
+    """
+    Runner for experiments with GANs
+    """
+
+    _default_experiment = GanExperiment
+    TRAIN_PARAMS = {
```
TezRomacH: can they be redefined?

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+        callbacks["phase_manager"] = PhaseManagerCallback(
+            train_phases=OrderedDict(
+                [
+                    (discriminator_phase, discriminator_phase_num),
+                    (generator_phase, generator_phase_num),
+                ]
+            ),
+            valid_mode="all",
+        )
+        callbacks["loss_g"] = PhaseBatchWrapperCallback(
```
TezRomacH: hm... i do not like this hardcoded strings.
let's make them the params for `prepare_callbacks`

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+        logdir: str = "./logs",
+        verbose: bool = False,
+        check: bool = False,
+    ):
+        # Validate models type and its items
+        self.validate_models(models)
+        # Check for optimizers
+        assert (
+            self.generator_key in optimizers.keys()
+        ), "optimizers must have optimizer with {self.generator_key} key"
```
TezRomacH: with `f-string`?

catalyst/utils/typing.py
TezRomacH: just use `Optimizer = optim.Optimizer`

catalyst/utils/typing.py
```diff
@@ -1,12 +1,23 @@
-from typing import Union  # isort:skip
+from typing import Union, Dict  # isort:skip
 
 import torch
 from torch import nn, optim
+from torch.utils import data
 
-Model = nn.Module
+Model = Union[nn.Module, Dict[str, nn.Module]]
```
TezRomacH: i'm not sure about this. This line makes the `Model` unclear. I would prefer to use `Union[nn.Module, Dict[str, nn.Module]]`  where it's needed explicitly

catalyst/utils/typing.py
```diff
@@ -1,12 +1,23 @@
...
 Criterion = nn.Module
-Optimizer = optim.Optimizer
+Optimizer = Union[optim.Adam, optim.SGD]
 Scheduler = optim.lr_scheduler._LRScheduler  # noinspection PyProtectedMember
+Dataset = data.Dataset
```
TezRomacH: this is very good :)

catalyst/utils/typing.py
```diff
@@ -1,12 +1,23 @@
...
+DataLoader = data.DataLoader
```
TezRomacH: i like this too :)

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+        )
+        callbacks["optim_d"] = PhaseBatchWrapperCallback(
+            base_callback=OptimizerCallback(
+                loss_key="loss_d", optimizer_key=self.discriminator_key,
+            ),
+            active_phases=[discriminator_phase],
+        )
+        return OrderedDict(callbacks)
+
+    def train(
```
TezRomacH, line 252: looks promising 👍 but we need to add docstrings

TezRomacH (changes requested):

catalyst/utils/typing.py
Arquestro: PyCharm showed to me that there is no such reference in torch.optim. Ill investigate this more.

catalyst/dl/experiment/gan.py
Arquestro: In typing.py I've set Model = Union[torch.nn.Module, Dict[str, torch.nn.Module]].
Yeah, good point to change that one too.

catalyst/utils/typing.py
Arquestro: Syntax sugar to make beautiful catalyst typing :)
Ill change it just give me a word. :D

catalyst/utils/typing.py
Arquestro: Probably i've been confused by it, so i'll return it back as it was.

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,122 @@
...
+from catalyst.utils.typing import (
+    Model,
+    Criterion,
+    Optimizer,
+    Scheduler,
+    Dataset
+)
+
+
+class GanExperiment(Experiment):
```
reyzharkov: A lot of functionality is the same as in `BaseExperiment`, maybe you should consider modifying `BaseExperiment` a little and define only necessary methods in `GANExperiment`

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,122 @@
...
+    def __init__(
+        self,
+        models: Union[Model, Dict[str, Model]],
+        criterions: Dict[str, Criterion],
+        optimizers: Dict[str, Optimizer],
+        schedulers: Dict[str, Scheduler] = {},
+        callbacks: "Dict[str, Callback]" = {},  # noqa: F821, E501
+        datasets: Dict[str, Any] = {},
```
reyzharkov: Why do you need datasets if you have loaders?

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def get_datasets(self, stage: str, **kwargs) -> OrderedDict[str, Dataset]:
+        return self._datasets.get(stage, None)
+
+    def get_loaders(self, stage: str) -> OrderedDict[str, DataLoader]:
+        return self._loaders.get(stage, None)
+
+    @staticmethod
+    def get_transforms(stage: str = None, mode: str = None):
+        return NotImplementedError(
```
reyzharkov: raise?

catalyst/dl/runner/gan.py
reyzharkov: TRAIN_STATE_PARAMS?

catalyst/dl/runner/gan.py
```diff
@@ -1,21 +1,59 @@
...
+    }
+
     def __init__(
         self,
-        model=None,
-        device=None,
-        features_key="features",
-        generator_key="generator",
-        discriminator_key="discriminator"
+        model: Model = None,
```
reyzharkov: Union[Model, Dict[Model]]?

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
             # --> g_loss (fake logits + REAL targets)
             return {
                 "fake_features": fake_features,  # visualization purposes only
-                "fake_logits": fake_logits
+                "fake_logits": fake_logits,
             }
         else:
             raise NotImplementedError(f"Unknown phase: self.state.phase")
 
+    def validate_models(self, models: Model) -> None:
```
reyzharkov: models: Dict[Model]?

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+        assert isinstance(
+            models, dict
+        ), "models must be of Dict[str, torch.nn.Module] type"
+        for key, model in models.items():
+            assert isinstance(
+                model, torch.nn.Module
```
reyzharkov: isinstance(model, Model)?

catalyst/dl/runner/gan.py
reyzharkov: +1, maybe add to __init__ train_state_params=None

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+            ), "model {key} must be torch.nn.Module type"
+        discriminator_assert_message = \
+            "models must have discriminator Module with {self.discriminator_key} key"  # noqa: E501
+        assert self.discriminator_key in models, discriminator_assert_message
+        assert isinstance(
```
reyzharkov: maybe it will be more convenient to have `model_generator=..., model_discriminator=...` in __init__?

catalyst/utils/typing.py
TezRomacH: i truly think, adding `Dataset` and `DataLoader` to our typing was a pretty good idea :)

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,122 @@
...
+    Dataset
+)
+
+
+class GanExperiment(Experiment):
+    def __init__(
+        self,
+        models: Union[Model, Dict[str, Model]],
+        criterions: Dict[str, Criterion],
+        optimizers: Dict[str, Optimizer],
```
reyzharkov: as criterion and optimizer may be key_value (even in regular experiment), I believe you should not write them in plural form

by the way, plural form for criterion is criteria=)

catalyst/dl/runner/gan.py
reyzharkov: oh, I see, the previous implementation was the same, so you can let it as it is now, but I hope in the future we should refactor this to be more convenient

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+
+    def train(
+        self,
+        models: Model,
+        criterion: Criterion,
+        optimizers: Dict[str, Optimizer],
+        datasets: Dict[str, Dataset],
+        loaders: Dict[str, DataLoader],
+        callbacks: Dict[str, Callback] = {},
+        discriminator_phase_num: int = 1,
```
reyzharkov: I believe it will be a better idea to require (or add the default) `PhaseCallback`. Let me explain this: there may be different phase manipulation strategies, e.g. train generator until discriminator is at most 75% right, train discriminator until it is at least 90% right. So in order to support more generic phase manipulation I propose to set directly PhaseCallback instead

catalyst/dl/runner/gan.py
reyzharkov: Isn't there a more beautiful way to determine this?
I can now propose these variants:
1. provide to `__init__` `generator_callbacks=[], discriminator_callbacks=[], shared_callbacks=[]` where each list contains `Callback`'s to be wrapped (and do not provide `callbacks=[]`)
2. provide to `__init__` 1)`callbacks=[...]` as it is right now and 2)`phase2callbacks=Dict[str, List[str]]` where the list contains the names(keys) of the activated callbacks

AND THIS SHOULD BE DONE IN EXPERIMENT CLASS, NOT IN RUNNER!

catalyst/utils/typing.py
reyzharkov: +1
Now I understand why you write `model: Model` everywhere, but same as Roman I do not think that redefine `Model` in such a way is a good idea

catalyst/dl/runner/gan.py
Arquestro: Haven't noticed or forgot about it. Thank you :)

catalyst/dl/runner/gan.py
Arquestro: They can be redefined as a default params for prepare_callbacks or in it's inner scope

catalyst/utils/typing.py
reyzharkov: I also like this, but are you sure that it will be clear for everyone? I mean, `torch.utils.data.Dataset` is pretty much obvious, but `catalyst.utils.typing.Dataset` from the first look (without looking how it was defined) seems like some catalyst-internal special dataset subclass. Will it not look confusing?

So, I like this, but I am afraid it may look confusing. But if you disagree with me it's ok, let it be

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+            models[self.discriminator_key], torch.nn.Module
+        ), discriminator_assert_message
+        generator_assert_message = \
+            "models must have generator Module with {self.generator_key} key"  # noqa: E501
+        assert self.generator_key in models, generator_assert_message
+        assert isinstance(
+            models[self.generator_key], torch.nn.Module
+        ), generator_assert_message
+
+    def prepare_state_params(
```
reyzharkov: move to experiment class?

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+        state_params = {
+            "num_epochs": num_epochs,
+            "logdir": logdir,
+            **self.TRAIN_PARAMS,
+        }
+        state_params.update(additional_state_params)
+        return state_params
+
+    @staticmethod
+    def prepare_experiment_params(
```
reyzharkov: move to experiment class?

catalyst/dl/runner/gan.py
```diff
@@ -61,18 +96,190 @@ def forward(self, batch):
...
+            "optimizers": {train_stage_name: optimizers},
+            "stages": [train_stage_name],
+            "state_params": {train_stage_name: state_params},
+            "callbacks": {train_stage_name: callbacks},
+            "datasets": {train_stage_name: datasets},
+            "loaders": {train_stage_name: loaders},
+        }
+        return experiment_params
+
+    def prepare_callbacks(
```
reyzharkov: move to experiment class?

reyzharkov (changes requested):

catalyst/dl/experiment/gan.py
Arquestro: I'll think about refactoring of BaseExperiment to inherit from it. Haven't done this for the sake of keeping neighbour pipelines untouched.

catalyst/dl/experiment/gan.py
Arquestro: To provide experiment interface with datasets. The only reason now. :)

catalyst/dl/runner/gan.py
Arquestro: Will move them to prepare_state_params later. They're supposed to be default and required.

catalyst/dl/runner/gan.py
Arquestro: Discussed with Roman, will make all these typing entries explicit.

catalyst/dl/experiment/gan.py
Arquestro: BaseExperiment is made as an one stage experiment. This means that every get_* accessor with stage returns inner fields as-is without addressing them as a dictionaries with stage keys.
I've thought about making BaseExperiment multi stage without breaking this behaviour.
Maybe add BaseOneStageExperiment on top of BaseExperiment, that is rewritten to be universal multistage experiment.

What do you think, @TezRomacH, @asmekal?

Arquestro: Or we can agree on that GanExperiment should be one staged for now, and just inherit it from BaseExperiment.
I don't like this choice :)

catalyst/utils/typing.py
Arquestro: I'll leave it and we'll see.
For me, it's not more confusing than Model that is nn.Module :)
I think we should do it one way or another for majority of typing entities.

catalyst/utils/typing.py
Arquestro: Reverted it

catalyst/dl/runner/gan.py
Arquestro: After transfer of callbacks processing into GanExperiment i'll dispose of hardcoded strings.

catalyst/dl/runner/gan.py
Arquestro: Changed to Dict[str, Model]. I suppose that there will be more than one model always.

catalyst/dl/experiment/gan.py
Arquestro: Will remove datasets, we don't need them, imo

Arquestro:
Based on the feedback i've reworked GanRunner train method and GanExperiment.
In this version GanExperiment is a subclass of BaseExperiment, and GanRunner does very little for experiment initialization, only validation of some parameters and addition to state_kwargs dict parameter. Still WIP.

Arquestro:
Unresolved all conversations.

catalyst/dl/runner/gan.py
```diff
@@ -29,25 +54,24 @@ def _prepare_for_stage(self, stage: str):
...
+            "generator_train_phase",
         ]:
-            assert hasattr(self.state, key) \
+            assert (
+                hasattr(self.state, key)
                 and getattr(self.state, key) is not None
+            )
 
-    def forward(self, batch):
+    def forward(self, batch, **kwargs):
```
TezRomacH: you pass `**kwargs` but don't use it 🤔

Arquestro: Will remove it

Scitator:
@Arquestro could you please merge PR with master branch?

Arquestro:
> @Arquestro could you please merge PR with master branch?

I'll do it today, but before this I need to revert some changes done by merge and add inference routine.

Arquestro:
@lightforever everything is alright with TeamCity?
image

lightforever:
@Arquestro, please merge master into your branch. Everything should be ok.

catalyst/core/runner.py
```diff
@@ -318,7 +318,11 @@ def _run_epoch(self, stage: str, epoch: int):
             self.state.loader_name = loader_name
             self.state.loader_len = len(loader)
             self.state.need_backward = loader_name.startswith("train")
-            self.model.train(self.state.need_backward)
+            if isinstance(self.model, dict):
+                for submodel in self.model:
```
TezRomacH:
```suggestion
                for model in self.model.values():
```
?

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,158 @@
...
+    PhaseManagerCallback, RaiseExceptionCallback, VerboseLogger
+)
+from catalyst.utils.tools.typing import Criterion, Model, Optimizer, Scheduler
+from .base import BaseExperiment
+
+
+class GanExperiment(BaseExperiment):
+    """
+    One-staged GAN experiment
+    Mainly used in notebooks
```
TezRomacH: Remove please `Mainly used in notebooks`

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+
+    def __init__(
+        self,
+        model: Union[Model, Dict[str, Model]],
```
TezRomacH: should it work if we pass only one model instead of a dict?

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,158 @@
...
+        loaders: "OrderedDict[str, DataLoader]",
+        callbacks: "OrderedDict[str, Callback]" = None,
```
TezRomacH: we also should support a `List[Callback]`, like in `SupervisedRunner.train`

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,158 @@
...
+        for phase_name, callback_name_list in phase2callbacks.items():
+            # TODO: Check for phase in state_params?
+            for callback_name in callback_name_list:
+                callback = self._callbacks.pop(callback_name)
+                self._callbacks[callback_name] = PhaseBatchWrapperCallback(
+                    base_callback=callback, active_phases=[phase_name],
+                )
+
+    def get_callbacks(self, stage: str) -> "OrderedDict[str, Callback]":
+        callbacks = super().get_callbacks(stage="")
```
TezRomacH: why is `stage=""`?

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,158 @@
...
+        for callback_name, callback_fn in default_callbacks:
+            is_already_present = any(
+                isinstance(x, callback_fn) for x in callbacks.values()
+            )
+            if not is_already_present:
+                callbacks[callback_name] = callback_fn()
+        return callbacks
+
+    def get_transforms(self, stage: str = None, dataset: str = None):
+        raise NotImplementedError(
```
TezRomacH: maybe just `pass`? 🤔 
what do you think @Scitator @bagxi @lightforever ?

catalyst/dl/runner/gan.py
```diff
@@ -243,5 +245,85 @@ def _discriminator_train_phase(self):
...
+        state_kwargs: Dict = None,
+        distributed_params: Dict = None,
+        monitoring_params: Dict = None,
+        logdir: str = None,
+        verbose: bool = False,
+        initial_seed: int = 42,
+        check: bool = False,
+    ) -> None:
+        """
+        :param model: models, usually generator and discriminator
```
TezRomacH: to google style docstring please

catalyst/dl/runner/gan.py
```diff
@@ -243,5 +245,85 @@ def _discriminator_train_phase(self):
...
+            for monitoring services
+        :param logdir: path to output directory
+        :param verbose: if true, it displays the status of the training
+            to the console.
+        :param initial_seed: experiment's initial seed value
+        :param check: if True, then only checks that pipeline is working
+            (3 epochs only)
+        """
+        # Check phase parameters in state_kwargs
+        consistent_metrics_param_key = "batch_consistant_metrics"
```
TezRomacH: -> init args?

TezRomacH (changes requested):

catalyst/core/runner.py
Arquestro: Agree. Pushed this change

catalyst/core/runner.py
```diff
@@ -319,8 +319,8 @@ def _run_epoch(self, stage: str, epoch: int):
             self.state.loader_len = len(loader)
             self.state.need_backward = loader_name.startswith("train")
             if isinstance(self.model, dict):
-                for submodel in self.model:
-                    self.model[submodel].train(self.state.need_backward)
+                for model in self.model.items():
```
TezRomacH: No, `items()` returns a tuple `(key, value)`

You need `values()`

TezRomacH (changes requested):

catalyst/dl/experiment/gan.py
Arquestro: No default param value for this argument, and it's not used in BaseExperiment, which is super class for GanExperiment

catalyst/core/runner.py
Arquestro: Thank you, fixed it

catalyst/dl/runner/gan.py
Arquestro: What do you mean?
This variable should be here, I guess, it's Notebook specific parameter key for training.

catalyst/dl/runner/gan.py
Arquestro: Aw, I followed @asmekal style for MultiphaseRunner and GanRunner init as the examples.
I guess i have to rewrite them too.

catalyst/dl/experiment/gan.py
Arquestro: Done

catalyst/dl/experiment/gan.py
Arquestro: No, dict intended nearly always. removed union with model part.

catalyst/dl/experiment/gan.py
Arquestro: Ok, but the main reason of GanExperiment is about callback wrapping, and it's strange if you wrap something without knowing how to address it(key). Added it to type hint anyway.

catalyst/dl/experiment/gan.py
Arquestro: Left it as it is now.

catalyst/core/runner.py
```diff
@@ -318,7 +318,11 @@ def _run_epoch(self, stage: str, epoch: int):
             self.state.loader_name = loader_name
             self.state.loader_len = len(loader)
             self.state.need_backward = loader_name.startswith("train")
-            self.model.train(self.state.need_backward)
+            if isinstance(self.model, dict):
```
reyzharkov: what about `maybe_recursive_call`?

catalyst/dl/experiment/gan.py
reyzharkov: But why not `stage=stage`? If it's not used right now it doesn't mean that it won't be used in the future

catalyst/dl/experiment/gan.py
reyzharkov: The error string description is probably misleading, @Arquestro what did you mean here?
"No transforms are used in GAN experiment" looks... clearly wrong as you may and should use them.

I suppose you can just do not implement `get_transforms` to force implementation in descendants or modify error message to make it clear why this method is not implemented

reyzharkov: @TezRomacH I think raising error is more safe here than `pass`

catalyst/core/runner.py
Arquestro: Thank you, tested it right now - works like a charm.

catalyst/dl/experiment/gan.py
Arquestro: Every pull request gives state for some functionality.
There are no subclasses of `GanExperiment` or routines that will use it's transforms getter, so raising `NotImplementedError` is right, because it's the part of GanExperiment state for following release.
What I agree about, is misguiding nature of the exceptions message, will change it.

catalyst/dl/experiment/gan.py
Arquestro: Will make it `stage=stage`

catalyst/core/runner.py
Scitator: do we really need such if in the core runner?
from Catalyst typing Model is instance of nn.Module... and maybe nn.ModuleDict, but not dict itself

catalyst/dl/experiment/base.py
```diff
@@ -13,9 +13,10 @@ class BaseExperiment(Experiment):
     Super-simple one-staged experiment
         you can use to declare experiment in code
     """
+
     def __init__(
         self,
-        model: Model,
+        model: Union[Model, Dict[str, Model]],
```
Scitator: no sure if we need such hack, why not to use nn.ModuleDict instead?

catalyst/dl/experiment/gan.py
Scitator: @Arquestro why do we need to raise an error here?

catalyst/dl/runner/gan.py
Scitator: what do you think about GanState with new default parameter? This look much more robust for possible future refactoring and methods renaming

catalyst/dl/experiment/gan.py
bagxi: @Arquestro could you please explain why do you need `get_transforms` now? seems you only re-raising exception with another message
maybe it is better to remove `get_transforms`, like in `SupervisedExperiment`?

catalyst/dl/runner/gan.py
Arquestro: My choice is to leave GanState for later.
I will remove this workaround for this default parameter. It will be user's responsibillity to define it in `state_kwargs`.

catalyst/dl/experiment/gan.py
Arquestro: @bagxi only to reraise it with another message, the only reason. And to remove PyCharms annoying warning about implementation of all abstract methods.

Arquestro: Removed this method.

catalyst/dl/experiment/base.py
Arquestro: I suggest to leave it as it is for now, for compatibility.

catalyst/dl/core/state.py
```diff
@@ -53,4 +52,49 @@ def __init__(
         )
 
 
-__all__ = ["State"]
+class GanState(_State):
```
Scitator: I don't think it's Core, it better to place at `catalyst/dl/state/gan.py`, just at SupervisedRunner

Scitator (changes requested):
+

catalyst/dl/runner/gan.py
```diff
@@ -2,8 +2,9 @@
     Any, Callable, Dict, List, Mapping, Optional, Tuple, Union
 )
 
-from catalyst.dl import Runner
-from catalyst.utils.tools.typing import Device, Model
+from catalyst.dl import GanExperiment, Runner
+from catalyst.dl.state import GanState
```
Scitator: why do you need such import? looks like `from catalyst.dl import GanExperiment, GanState , Runner` should be enough

catalyst/dl/state/__init__.py
```diff
@@ -0,0 +1,4 @@
+# flake8: noqa
+# isort:skip_file
+
+from .gan import GanState
```
Scitator, line 3: also should be imported in catalyst/dl/__init__

catalyst/dl/core/state.py
```diff
@@ -30,7 +30,6 @@ def __init__(
         batch_consistant_metrics: bool = True,
         **kwargs
     ):
-
         # @TODO: remove GAN hack
         self.phase = None
```
Scitator: let's add phase into GanState, rather then typical DLState

Scitator:
btw, @Arquestro could you please add some CI test, that everything import correctly and you can create instances for new classes?
something like https://github.com/catalyst-team/catalyst/blob/master/catalyst/contrib/nn/tests/test_optimizer.py

catalyst/dl/runner/gan.py
```diff
@@ -76,7 +77,11 @@ class GanRunner(MultiPhaseRunner):
     Various conditioning types, penalties and regularization (such as WGAN-GP)
     can be easily derived from this class
     """
-    _default_experiment = GanExperiment
+    experiment: GanExperiment
```
reyzharkov, line 79: why not just `_default_experiment`?  And all of the rest - where are you using them?

Scitator: _default_experiment is used to create "new" Runners for new Experiments, during .run_experiment, or something like that customisation
the idea is quite simple - Experiments have the same API, so you don't need to rewrite whole .train method, only on class parameter

reyzharkov: ok, seems like I didn't notice refactoring in core

catalyst/dl/core/state.py
Arquestro: Done

catalyst/dl/runner/gan.py
Arquestro: Done

catalyst/dl/state/__init__.py
Arquestro: Done

catalyst/dl/core/state.py
Arquestro: Done

catalyst/dl/experiment/gan.py
```diff
@@ -37,7 +37,7 @@ def __init__(
         distributed_params: Dict = None,
         monitoring_params: Dict = None,
         initial_seed: int = 42,
-        phase2callbacks: Dict[str, List[str]] = None,
+        phase2callbacks: Dict[str, List[str]] = {},
```
Scitator: it's better to use None and `phase2callbacks  or {}` after that

Scitator (commented):
+

catalyst/dl/experiment/gan.py
Arquestro: Done

catalyst/dl/state/gan.py
```diff
@@ -0,0 +1,55 @@
...
+        logdir: str = None,
+        stage: str = "infer",
+        num_epochs: int = 1,
+        main_metric: str = "loss",
+        minimize_metric: bool = True,
+        valid_loader: str = "valid",
+        verbose: bool = False,
+        checkpoint_data: Dict = None,
+        batch_consistant_metrics: bool = False,
+        **kwargs
```
Scitator: let's use 
```
batch_consistant_metrics: bool = False,
**kwargs
```
?

Scitator (commented):
+

catalyst/dl/experiment/gan.py
Scitator: @Arquestro what do you think about 
```
phase2callbacks: Dict[str, List[str]] = None,
**kwargs
```

Scitator (commented):
+

catalyst/dl/state/gan.py
Arquestro: Done

catalyst/dl/experiment/gan.py
```diff
@@ -0,0 +1,114 @@
...
+class GanExperiment(BaseExperiment):
+    """
+    One-staged GAN experiment
+    """
+
+    def __init__(
+        self,
+        *,
+        phase2callbacks: Dict[str, List[str]] = None,
+        **kwargs,
```
TezRomacH: let's put all the params with types 🤔

TezRomacH (approved):

PR #595, opened 2020-01-16
ConfigAPI add transform_params

bagxi:
## Type of Change

## Checklist


catalyst/utils/image.py
```diff
@@ -110,8 +107,7 @@ def mimread(
         uri = uri if uri.startswith(rootpath) else os.path.join(rootpath, uri)
 
     image = np.dstack(imageio.mimread(uri, **kwargs))
-    if clip_range is not None:
-        image = np.clip(image, *clip_range)
+    image = np.clip(image, 0, 1).astype(np.float32)
```
Scitator: why did you make clip_range by default?

catalyst/dl/experiment/config.py
```diff
@@ -319,6 +321,35 @@ def get_scheduler(self, stage: str, optimizer: Optimizer) -> Scheduler:
...
+                for transform_params in params["transforms"]
+            ]
+            params.update(transforms=transforms_composition)
+
+        transform = TRANSFORMS.get_from_params(**params)
+
+        return transform
+
+    def get_transforms(self, stage: str = None, mode: str = None) -> Callable:
+        transform_params = deepcopy(
```
Scitator, line 367: 🤔 could you please add `_key_value` logic, so we could sync the transformers for different key by [Augmentor](https://github.com/catalyst-team/catalyst/blob/master/catalyst/data/augmentor.py#L4)?

Scitator: moreover, we could also add TupleAugmentor to handle tuple-like bathes, like MNIST and CIFAR one

catalyst/utils/image.py
bagxi: do you mean that it is better to add `clip_range` parameter to `MaskReader` rather than making it by default?

Scitator: yup, it some internal logic and it's better to add params to be able to change it

catalyst/dl/experiment/config.py
bagxi: I'm not sure I follow you, could you please provide an example of `_key_value` for `transform_params`?

do you mean something like this
```
transform_params:
  _key_value: True

  image:
    transform: Compose
    transforms:
      - transform: SomeTransform
      - ...
  mask:
    transform: Compose
    transforms:
      - ...
  keypoint:
    transform: ...
```
?

Scitator: yes, just like that

catalyst/data/augmentor.py
```diff
@@ -6,22 +6,42 @@ class Augmentor:
     Augmentation abstraction to use with data dictionaries.
     """
     def __init__(
-        self, dict_key: str, augment_fn: Callable, default_kwargs: Dict = None
+        self,
+        dict_key: str,
+        augment_fn: Callable,
+        input_key: str = None,
+        output_key: str = None,
+        default_kwargs: Dict = None,
```
Scitator: I think, you can use `**default_kwargs`

catalyst/dl/registry.py
```diff
@@ -14,8 +19,39 @@ def _callbacks_loader(r: Registry):
...
+    except ImportError as ex:
+        logger.warning(
+            "albumentations not available, to install albumentations, "
+            "run `pip install albumentations`."
+        )
+        if os.environ.get("USE_ALBUMENTATIONS", "0") == "1":
+            raise ex
+
+
+TRANSFORMS = Registry("transform")
```
Scitator, line 36: maybe, `"transforms"`?

catalyst/dl/experiment/config.py
```diff
@@ -319,6 +322,66 @@ def get_scheduler(self, stage: str, optimizer: Optimizer) -> Scheduler:
...
+                    ConfigExperiment._get_transform(**transform_params)
+                    for transform_params in params["transforms"]
+                ]
+                params.update(transforms=transforms_composition)
+
+            transform = TRANSFORMS.get_from_params(**params)
+
+        return transform
+
+    def get_transforms(self, stage: str = None, mode: str = None) -> Callable:
```
Scitator: I am not sure it's a `mode`, from your examples it looks like `loader_{name}`/`dataset_{name}`/ etc?

catalyst/dl/registry.py
bagxi: I think, no
we use `"transforms"` to define a list of transforms inside transform (e.g. for `albumentations.Compose`) and as registry provides factory by name it will lead to key duplication, e.g.:
```
transforms: Compose
transforms:
  - transform: Normalize
  - transform: ToTensorV2
```

catalyst/data/augmentor.py
bagxi: I'm not sure I follow you, do you mean that it is better to replace `default_kwargs: Dict = None,` with `**default_kwargs`?

catalyst/dl/experiment/config.py
bagxi: yes, what name instead of `mode` could you suggest to use to prevent confusion?

catalyst/data/augmentor.py
Scitator: yup, I think it's better for user experience to just write `(... output_key=str, some_val=1)` rather than `(... output_key=str, default_kwargs={some_val=1})`

catalyst/dl/experiment/config.py
Scitator: what do you think about `loader` (as we are using transforms for loaders)

catalyst/dl/registry.py
Scitator: thinking, 
`registry.transform(MyTransform)` vs `registry.transforms(MyTransform)`

catalyst/dl/experiment/config.py
```diff
@@ -323,32 +324,63 @@ def get_scheduler(self, stage: str, optimizer: Optimizer) -> Scheduler:
...
-        if transform is not None:
-            def process(dict_):
-                result = transform(**dict_)
-                return result
-        else:
-            def process(dict_):
+        if transform is None:
+            def transform(dict_):
                 return dict_
+        elif not isinstance(transform, AugmentorCompose):
```
baibaidj, line 379: I think this line is a bug. Using the transform.yml provided in _test_cv_segmentation, the resulting transform before this line is a Compose object and will be judged as NOT AugmentorCompose. Finally the returning transform cannot accept the dict with 'image' and 'mask' keys as input, thus leading to error. Without this judgment (e.g. removing the not), the transform will behave as anticipated.

Scitator: could you please make an issue?
here is the same code - https://github.com/catalyst-team/catalyst/blob/master/catalyst/dl/experiment/config.py#L415

PR #600, opened 2020-01-21
SMP init

TezRomacH:
## Description

- Added `prefix` parameter for registry from a module. Example `r.add_from_module(module, prefix="module.")`
- Added support for [SMP](https://github.com/qubvel/segmentation_models.pytorch) models to the Config API
- hotfix IoU with no reduce dim
- Added test example for segmentation

## Type of Change

## Checklist

### DIFF
--- a/catalyst/contrib/transforms/tensor.py
+++ b/catalyst/contrib/transforms/tensor.py
@@ -5,19 +5,38 @@ import torch
 
 
 class TensorToImage(ImageOnlyTransform):
+    """
+    Casts torch.tensor to numpy array
+    """
     def __init__(self, always_apply=False, p=1.0):
+        """
+        Args:
+            always_apply (bool): need to apply this transform anyway
+            p (float): probability for this transform
+        """
         super().__init__(always_apply, p)
 
     def apply(self, img, **params):
+        """Apply the transform to the image"""
         return np.expand_dims(img.numpy(), axis=0)
 
 
 class ToTensor(ImageOnlyTransform):
+    """
+    Casts numpy array to torch.tensor, but do not move channels dim
+    """
     def __init__(self, always_apply=False, p=1.0):
+        """
+        Args:
+            always_apply (bool): need to apply this transform anyway
+            p (float): probability for this transform
+        """
         super().__init__(always_apply, p)
 
     def apply(self, img, **params):
+        """Apply the transform to the image"""
         return torch.from_numpy(img)
+
 
 __all__ = [
     "TensorToImage",
--- a/catalyst/contrib/transforms/tensor.py
+++ b/catalyst/contrib/transforms/tensor.py
@@ -18,3 +18,8 @@ class ToTensor(ImageOnlyTransform):
 
     def apply(self, img, **params):
         return torch.from_numpy(img)
+
+__all__ = [
+    "TensorToImage",
+    "ToTensor"
+]
--- a/catalyst/contrib/transforms/tensor.py
+++ b/catalyst/contrib/transforms/tensor.py
@@ -1,6 +1,7 @@
 import numpy as np
 
-from albumentations.core.transforms_interface import ImageOnlyTransform
+from albumentations import ImageOnlyTransform
+from albumentations.pytorch import ToTensorV2
 import torch
 
 
@@ -21,21 +22,41 @@ class TensorToImage(ImageOnlyTransform):
         return np.expand_dims(img.numpy(), axis=0)
 
 
-class ToTensor(ImageOnlyTransform):
+class ToTensor(ToTensorV2):
     """
-    Casts numpy array to torch.tensor, but do not move channels dim
+    Casts numpy array to ``torch.tensor``
     """
-    def __init__(self, always_apply=False, p=1.0):
+    def __init__(
+        self,
+        move_channels_dim: bool = True,
+        always_apply: bool = False,
+        p: float = 1.0,
+    ):
         """
         Args:
+            move_channels_dim (bool): if ``False``, casts numpy array
+                to ``torch.tensor``, but do not move channels dim
             always_apply (bool): need to apply this transform anyway
             p (float): probability for this transform
         """
         super().__init__(always_apply, p)
+        self.move_channels_dim = move_channels_dim
 
     def apply(self, img, **params):
         """Apply the transform to the image"""
-        return torch.from_numpy(img)
+        if self.move_channels_dim:
+            return super().apply(img, **params)
+        else:
+            return torch.from_numpy(img)
+
+    def apply_to_mask(self, mask: np.ndarray, **params) -> torch.Tensor:
+        """Apply the transform to the mask"""
+        if self.move_channels_dim:
+            mask = mask.transpose(2, 0, 1)
+        return super().apply_to_mask(mask.astype(np.float32), **params)
+
+    def get_transform_init_args_names(self):
+        return ("move_channels_dim",)
 
 
 __all__ = [
--- a/catalyst/contrib/transforms/tensor.py
+++ b/catalyst/contrib/transforms/tensor.py
@@ -4,22 +4,43 @@ from albumentations import ImageOnlyTransform
 from albumentations.pytorch import ToTensorV2
 import torch
 
+from catalyst.utils.image import tensor_to_ndimage
+
 
 class TensorToImage(ImageOnlyTransform):
     """
     Casts torch.tensor to numpy array
     """
-    def __init__(self, always_apply=False, p=1.0):
+    def __init__(
+        self,
+        denormalize: bool = False,
+        move_channels_dim: bool = True,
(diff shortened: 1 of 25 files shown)


Scitator:
@TezRomacH 
Hi, could you please merge with master and add some tests for modules import and their Config API usage?

catalyst/contrib/transforms/tensor.py
```diff
@@ -0,0 +1,44 @@
+import numpy as np
+
+from albumentations.core.transforms_interface import ImageOnlyTransform
+import torch
+
+
+class TensorToImage(ImageOnlyTransform):
```
Scitator, line 10: why do we need this transforms here?
@bagxi what do you think?

catalyst/dl/registry.py
```diff
@@ -24,6 +24,12 @@ def _transforms_loader(r: Registry):
     try:
         import albumentations as m
         r.add_from_module(m)
+
+        from albumentations import pytorch as p
+        r.add_from_module(p)
```
Scitator: what to you think about adding `albu` prefix?

Scitator: or, maybe `A` for Albumentations and `C` for Catalyst?

catalyst/contrib/transforms/tensor.py
Scitator: btw, we also need to merge this PR first
https://github.com/catalyst-team/catalyst/pull/604

TezRomacH: This transformation is needed when torch.Tensor is returned from the dataset, not numpy. For example, MNIST in our tests

catalyst/dl/registry.py
TezRomacH: We do not support any other augmentation library (I think, we won't).

And I think it's unnecessary to add a prefix to all transforms in config

TezRomacH: About, `C` for Catalyst. Users will not use our contrib transforms very often, I prefer to use full form, it is much readable and clean

catalyst/contrib/transforms/tensor.py
Scitator: maybe, it should be `TensorToNumpy` then?

Scitator: or just `ToNumpy`?

catalyst/dl/registry.py
Scitator: there are 2 ways:
1. we add prefix for all our sources for transformations, except Core Catalyst one
2. we don't use any prefixes at all... but it could cause name collision
all other variants looks unscalable

PS. And yup, we will support more transformations source in the future

TezRomacH: > we add a prefix for all our sources for transformations, except Core Catalyst one
why is that? `albumentations` has a lot more common transforms than in catalyst.core

it's totally okay to use prefixes for some unusual classes or methods. Catalyst has only two transforms, and maybe in the future, it will have a few more. but albu has a lot of common transformations

TezRomacH: > PS. And yup, we will support more transformations source in the future

nice, and for them, i would favor adding a prefix, like `somenewlib.Transforms`

but making `albumentations` preferred one, without prefix

TezRomacH: look at a typical transform config
https://github.com/catalyst-team/catalyst/pull/600/files#diff-e8e0a33e0d5dfbfe552d3a0fc812d1bdR5

adding an `albu.` prefix for every single transform is long and unnecessary

catalyst/contrib/transforms/tensor.py
TezRomacH: it's okay to me, but this transform was already in catalyst.

so we need @bagxi  to discuss

bagxi: maybe, `TensorToImage`?

bagxi: and would it be useful if we will add parameter like `use catalyst.utils.tensor_to_ndimage`?

TezRomacH: let's make `tensor_to_ndimage` more general with flags like `need_normalization` and `need_move_channels_dim` and then use this function inside of `TensorToImage`/`ToNumpy`/`YouNameIt` transfrom

Scitator: prosed simplification, `normalize=`, `move_channels_dim=`

catalyst/dl/registry.py
Scitator: `but making albumentations preferred one, without prefix` why?

TezRomacH: ODS support

TezRomacH: the main point of this PR is a friendship between Catalyst, albumentations, SMP and maybe ttach in future

catalyst/contrib/transforms/tensor.py
bagxi: maybe it is better to use `denormalize=` (instead of `normalize=`) as in `utils.tensor_to_ndimage` we are multiplying image by std and adding mean?

Scitator: sure

catalyst/dl/registry.py
Scitator: I dont want to have any possible bias to any other framework (except Pytorch, of course),
I like ODS, nevertheless, we need to be open for any other framework on the same rules.

Scitator: you can use any prefix, even `A` is good enough, 
or add some global prefix-source for all transformations

TezRomacH: ok, i've added `A.` prefix

examples/_tests_mnist_stages2/config1.yml
```diff
@@ -17,11 +17,11 @@ stages:
   transform_params:
     transform: A.Compose
     transforms:
-      - transform: catalsyt.TensorToImage
+      - transform: C.TensorToImage
       - transform: A.Normalize
         mean: [0.1307]
         std: [0.3081]
-      - transform: catalsyt.ToTensor
+      - transform: C.ToTensor
```
TezRomacH: but why?

we have never called our framework `C`. using `catalsyt.ToTensor` is much clearer

TezRomacH: @Scitator

Scitator: first of all - **C**ata**lyst** :)
but, `Catalyst.ToTensor` looks too long and complicated

TezRomacH: `from catalyst import utils`

it's lowercase :)

and here: `transform: catalsyt.ToTensor` it means to use  `ToTensor` from sort of package `catalyst.`

TezRomacH: `C.ToTensor` is unclear and more complicated

Scitator: why do you think so? why is it unclear and more complicated?

TezRomacH: because, you will never know, what is `C.`, where else it's used

`smp.Unet` means you use Unet from module `segmentation_models_pytorch` also known as `smp`

`A.RandomCrop` means you use RandomCrop from `albumentations` also known as `A`

what does `C.ToTensor` mean?

TezRomacH: To peacefully end the argue, I suggest adding functionality that allows you to add several different prefixes for the same method or class.

For example, `.add_from_module(t, prefix=["catalyst.", "C."])`.

Scitator: you know `albumentaions as A`, nevertheless from my perspective they are `albu` or `albumentations `... it's a matter of taste.

but the idea with  different prefixes looks good, you are right :)

Scitator: you can also add aliases for albumentations, like `albumentations`, `albu`, `A`

TezRomacH: albumentations uses `A.` in their first and examples :)

TezRomacH: * in their docs

Scitator: would you like to make changes during this PR, or is it okay for you to make them on https://github.com/catalyst-team/catalyst/pull/607 ?
GANs are waiting for you, @TezRomacH :)

TezRomacH: I am literally working on this right now

give me 30-40 minutes

TezRomacH: done

TezRomacH:
build is done successfully, @Scitator

PR #618, opened 2020-01-30
Make Catalyst interesting for the industry!

cgarciae:
This change might be insignificant in the grand scheme of things, but the vast majority of DL applications are probably for doing "simple" stuff in production. Unless Catalyst is truly only for research, it wouldn't be harmful to be inclusive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 </div>
 
-PyTorch framework for DL & RL research.
+PyTorch framework for DL & RL research and development.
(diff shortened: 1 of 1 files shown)


Scitator:
Thanks, you are right,
Catalyst.Ecosystem is truly focused on DL R&D best practices,
https://docs.google.com/presentation/d/1D-yhVOg6OXzjo9K_-IS5vSHLPIUxp1PEkFGnpRcNCNU/edit?usp=sharing

PR #628, opened 2020-02-01
Fix/resume

### LINKED ISSUE #622: Repair resume: resume epoch and stage(!)
This bug is caused by the following reasons:
1. Checkpoint doesn't contain  `stage_epoch` and `Runner_ ` hasn't property information for resuming.
2. `Runner_` always starts stage from `stage_epoch=0`
3. `Runner_` doesn't check resumed stage.

AndreySheka:
## Description
fix #622 

## Type of Change

## Checklist

- [x] I have read I need to click 'Login as guest' to see Teamcity build logs


catalyst/core/callbacks/checkpoint.py
```diff
@@ -107,22 +107,26 @@ def get_checkpoint_suffix(self, checkpoint: dict) -> str:
...
-                model=state.model,
-                criterion=state.criterion,
-                optimizer=state.optimizer,
-                scheduler=state.scheduler
-            )
-
-            print(
-                f"loaded checkpoint {filename} (epoch {checkpoint['epoch']})"
-            )
+            if checkpoint["epoch"] > state.epoch:
```
TezRomacH: what is this for?

AndreySheka: When you resume on stage1, on stage2 you will resume again. This `if` helps to solve the problem.

Scitator: @AndreySheka could you please add a comment about this in the code?
or could we add some additional logic to check if we need to resume checkpoint loading again?
maybe add `self.resume = None` in the end of each stage?

this if looks really unclear and hacky for me, that the case :)

AndreySheka: I remove this hack, now checkpoint callback clear `state.resume` after loading checkpoint and `resume` field added into migration list of `runner`

catalyst/core/runner.py
```diff
@@ -350,8 +350,21 @@ def _run_stage(self, stage: str):
                 break
 
             self.state.epoch += 1
+            self.state.stage_epoch += 1
         self._run_event("stage", moment="end")
 
+    def get_start_stage(self):
+        resume, resume_dir = [getattr(self.state, key, None)
```
Scitator: there is now state before stage starts

Scitator (changes requested):

catalyst/core/runner.py
AndreySheka: fixed

catalyst/core/runner.py
```diff
@@ -350,8 +351,24 @@ def _run_stage(self, stage: str):
                 break
 
             self.state.epoch += 1
+            self.state.stage_epoch += 1
         self._run_event("stage", moment="end")
 
+    def get_start_stage(self):
+        state_params = self.experiment.get_state_params(
```
Scitator: maybe, it's better to add something like `resume` to Experiment to prevent all this hack around stages and state?

Scitator (changes requested):

catalyst/core/runner.py
AndreySheka: I think it should be placed in `Runner`, because `Runner` has full information about current learning process, `Runner` emit events and decide, which `stage` will calculate. In first vestion `get_start_stage` was placed after `_prepare_for_stage`. But in this case it looks like a hack, because you need to write something like this:
```
def _run_stage(self, stage: str):
        self._prepare_for_stage(stage)
        resume_stage = self.get_resume_stage()
        stages = list(self.experiment.stages)
        if stages.index(resume_stage) > stages.index(stage):
            return
```

Scitator: as the Runner is mostly about *how* to run the experiment, and experiments is mostly about *what* to run, I think it looks okay to place initial `resume` into experiment. 
maybe, with @gazay abstraction for ConfigParser, that will create appropriate Experiment.Config from any user config this hypothesis will also work too

AndreySheka: As for me saving and restoring is the part of _how_, because Runner can compute different stages with full restarting on each epoch. Now `resume` placed in `checkpoint` callback. As for me, using callback for resuming looks like a hack. Changes in this PR  focused on minimal changes, which will fix the bug. But now it looks like refactoring of architecture. 

I think resuming can be implemented in Experiment constructor, which returns cutted list of stages.  This approach is more intuitive. But in this case, restoring must be removed from checkpointcallback.

Scitator: yes, maybe checkpoint callback for restoing should be implemented in another logic, 
nevertheless, it the story for another PR
for now I think, that start stage logic should be on the Experiment side (the hack with runner is my main concern)

AndreySheka: I move get_start_stage from Runner to ConfigExperiment.

catalyst/dl/experiment/config.py
```diff
@@ -537,5 +538,23 @@ def get_callbacks(self, stage: str) -> "OrderedDict[Callback]":
...
+        resume, resume_dir = [state_params.get(key, None)
+                              for key in ["resume", "resume_dir"]]
+
+        if resume_dir is not None:
+            resume = resume_dir / str(resume)
+
+        if resume is not None and Path(resume).is_file():
+            checkpoint = utils.load_checkpoint(resume)
+            return checkpoint["stage"]
+        return self.stages[0]
```
Scitator: this hack will not work, because we need such logic in every Experiment (not only Config one)

could you please change the logic for stages property for ConfigExperiment?
github.com/catalyst-team/catalyst/blob/master/catalyst/dl/experiment/config.py#L116#L119
this looks like a more elegant solution, with any changes in Runner logic

Scitator (changes requested):

catalyst/dl/experiment/config.py
Scitator: @AndreySheka

AndreySheka: @Scitator I moved get_start_stage logic into stages property and rebased this branch to master. It's ready for merge.

PR #647, opened 2020-02-07
[Feature] Allow next stage to run from scratch

reyzharkov:
## Description

Allows next stage to run from scratch (epoch 0, step 0), which is useful if the training/validation dataset is changing through time (metrics incompatible), e.g. we trained on synthetic data and then finetune on real data.

```yaml
stage1:
  ...
stage2:
  state_params:
    migrate: False  # (default True)
```

## Type of Change

## Checklist

- [x] I have read I need to click 'Login as guest' to see Teamcity build logs


catalyst/core/runner.py
```diff
@@ -154,7 +154,8 @@ def _get_experiment_components(
     def _prepare_for_stage(self, stage: str):
         utils.set_global_seed(self.experiment.initial_seed)
         migrating_params = {}
-        if self.state is not None:
+        migrate = self.experiment.get_state_params(stage).get("migrate", True)
+        if self.state is not None and migrate:
```
Scitator: could you please create something like `stage_state_params`, and reuse them during State creation? I just don't like the idea for several `self.experiment.get_state_params` call, if there is no need in it

Scitator (changes requested):

catalyst/core/runner.py
```diff
@@ -154,7 +154,8 @@ def _get_experiment_components(
     def _prepare_for_stage(self, stage: str):
         utils.set_global_seed(self.experiment.initial_seed)
         migrating_params = {}
-        migrate = self.experiment.get_state_params(stage).get("migrate", True)
+        stage_state_params = self.experiment.get_state_params(stage)
+        migrate = stage_state_params.get("migrate", True)
```
Scitator: what do you think about `.pop` usage?

reyzharkov: maybe `migrate` will be important for some callbacks, so let it be?

Scitator: 🤔 what about naming then? maybe `migrate_stage=[True/False]`?

reyzharkov: You mean to rename `migrate` -> `migrate_stage`?

Scitator: @asmekal yup, to make it easier to understand

reyzharkov: ok, renamed to `migrate_from_previous_stage`

PR #667, opened 2020-02-17
Fix model tracing exception

valyukov:
## Description

`catalyst-dl trace` throw an exception when Runner has been configured to use output_keys

## Related Issue

none

## Type of Change

## Checklist

### DIFF
--- a/catalyst/dl/utils/trace.py
+++ b/catalyst/dl/utils/trace.py
@@ -45,6 +45,7 @@ class _TracingModelWrapper(nn.Module):
         self.tracing_result = torch.jit.trace(
             method_model, *args, **kwargs
         )
+        return self.model.forward(*args, **kwargs)
 
 
 def trace_model(


Scitator:
I see an issue with such solution, for example
```
model:
  forward:
    return dict
  forward_tuple
    return tuple
```
If you would like to trace `forward_tuple` your will get an error.

Scitator:
Could you please rewrite the solution to return typical `model.forward` as a result?

valyukov:
Could you elaborate which exactly error you expect to get?

BTW, if `model.forward` return `dict` it definitely lead to error, because current implementation of Runner can't deal with it properly.

Scitator:
@valyukov with current tracing implementations, you rewrite  model`.forward` method, so our model [output postprocessing](https://github.com/catalyst-team/catalyst/blob/master/catalyst/dl/runner/supervised.py#L57L62) becomes invalid 
for example, you model always returns dict, but your would like to trace method that returns tuple

PR #685, opened 2020-03-03
Fix wandb wrong step problem (Fix #684)

### LINKED ISSUE #684: SupervisedWandbRunner logs wrong number of epochs to WandB
**Describe the bug**
Catalyst 20.02.4
`WandbRunner` is logging wrong number of epochs to WandB

**To Reproduce**
```
from catalyst import dl
import torch
import torch.nn as nn
import torch.optim as optim
from torch.utils.data import DataLoader, TensorDataset

class Projector(nn.Module):
    def __init__(self, input_size):
        super().__init__()
        self.linear = nn.Linear(input_size, 1)
        
    def forward(self, X):
        return self.linear(X).squeeze(-1)

X = torch.rand(16, 10)
y = torch.rand(X.shape[0])
model = Projector(X.shape[1])
dataset = TensorDataset(X, y)
loader = DataLoader(dataset, batch_size=8)
runner = dl.SupervisedWandbRunner()

runner.train(
    model=model,
    loaders={
        "train": loader,
        "valid": loader
    },
    criterion=nn.MSELoss(),
    optimizer=optim.Adam(model.parameters()),
    logdir="log_xxx_000",
    monitoring_params={
        "project": "wandb_wrong_epochs"
    },
    num_epochs=10
)
```

**Expected behavior**
In WandB I see two plots with `MSELoss` with exactly 10 epochs

**Actual behaviour**
In WandB I see two plots with `MSELoss` with 20 epochs (but each has exactly 10 dots)

**Screenshots**
Look on number of steps. It has 20 steps. But should have 10.
image

CrafterKolyan:
## Description

Exactly 10 steps from 0 to 9.
image

## Related Issue

https://github.com/catalyst-team/catalyst/issues/684

## Type of Change

## Checklist

- [x] I have read I need to click 'Login as guest' to see Teamcity build logs.

### DIFF
--- a/catalyst/contrib/dl/runner/wandb.py
+++ b/catalyst/contrib/dl/runner/wandb.py
@@ -149,12 +149,18 @@ class WandbRunner(Runner):
             wandb.log(commit=True)
 
     def run_experiment(self, experiment: Experiment):
+        """Starts experiment
+        
+        Args:
+            experiment (Experiment): experiment class
+        """
         self._pre_experiment_hook(experiment=experiment)
         super().run_experiment(experiment=experiment)
         self._post_experiment_hook(experiment=experiment)
 
 
 class SupervisedWandbRunner(WandbRunner, SupervisedRunner):
+    """SupervisedRunner with WandB"""
     pass
 
 
--- a/catalyst/contrib/dl/runner/wandb.py
+++ b/catalyst/contrib/dl/runner/wandb.py
@@ -15,7 +15,9 @@ class WandbRunner(Runner):
     Runner wrapper with wandb integration hooks.
     """
     @staticmethod
-    def _log_metrics(metrics: Dict, mode: str, suffix: str = "", epoch: int = None):
+    def _log_metrics(
+        metrics: Dict, mode: str, suffix: str = "", epoch: int = None
+    ):
         def key_locate(key: str):
             """
             Wandb uses first symbol _ for it service purposes
@@ -136,7 +138,10 @@ class WandbRunner(Runner):
             )
             for mode, metrics in mode_metrics.items():
                 self._log_metrics(
-                    metrics=metrics, mode=mode, suffix=self.epoch_log_suffix, epoch=epoch
+                    metrics=metrics,
+                    mode=mode,
+                    suffix=self.epoch_log_suffix,
+                    epoch=epoch
                 )
             wandb.log(commit=True)
 
--- a/catalyst/contrib/dl/runner/wandb.py
+++ b/catalyst/contrib/dl/runner/wandb.py
@@ -16,7 +16,7 @@ class WandbRunner(Runner):
     """
     @staticmethod
     def _log_metrics(
-        metrics: Dict, mode: str, suffix: str = "", epoch: int = None
+        metrics: Dict, mode: str, suffix: str = "", commit: bool = True
     ):
         def key_locate(key: str):
             """
@@ -36,7 +36,7 @@ class WandbRunner(Runner):
             f"{key_locate(key)}/{mode}{suffix}": value
             for key, value in metrics.items()
         }
-        wandb.log(metrics, step=epoch)
+        wandb.log(metrics, commit=commit)
 
     def _init(
         self,
@@ -125,7 +125,10 @@ class WandbRunner(Runner):
             mode = self.state.loader_name
             metrics = self.state.batch_metrics
             self._log_metrics(
-                metrics=metrics, mode=mode, suffix=self.batch_log_suffix
+                metrics=metrics,
+                mode=mode,
+                suffix=self.batch_log_suffix,
+                commit=True
             )
 
     def _run_epoch(self, stage: str, epoch: int):
@@ -141,7 +144,7 @@ class WandbRunner(Runner):
                     metrics=metrics,
                     mode=mode,
                     suffix=self.epoch_log_suffix,
-                    epoch=epoch
+                    commit=False
                 )
             wandb.log(commit=True)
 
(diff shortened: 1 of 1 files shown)


catalyst/contrib/dl/runner/wandb.py
```diff
@@ -16,7 +16,9 @@ class WandbRunner(Runner):
     Runner wrapper with wandb integration hooks.
     """
     @staticmethod
-    def _log_metrics(metrics: Dict, mode: str, suffix: str = ""):
+    def _log_metrics(
+        metrics: Dict, mode: str, suffix: str = "", epoch: int = None
+    ):
```
Scitator: `_log_metrics` is used for both batch and epoch metrics, so `epoch: int = None` looks a bit confusing in this way

catalyst/contrib/dl/runner/wandb.py
```diff
@@ -129,8 +131,12 @@ def _run_epoch(self, stage: str, epoch: int):
             for mode, metrics in \
                     self.state.metric_manager.epoch_values.items():
                 self._log_metrics(
-                    metrics=metrics, mode=mode, suffix=self.epoch_log_suffix
+                    metrics=metrics,
+                    mode=mode,
+                    suffix=self.epoch_log_suffix,
+                    epoch=epoch
                 )
+            wandb.log(commit=True)
```
Scitator, line 149: why do we need to do this?

Scitator (changes requested):

catalyst/contrib/dl/runner/wandb.py
CrafterKolyan: From WandB documentation:
> As long as you keep passing the same value for `step`, W&B will collect the keys and values from each call in one unified dictionary. As soon you call `wandb.log()` with a different value for `step` than the previous one, W&B will write all the collected keys and values to the history, and start collection over again.

(https://docs.wandb.com/library/log)

This means that if there's no this line:
1. Results of epoch `i` are sent to WandB when you already obtained results for epoch `i + 1`
2. Results of last epoch are not sent because you won't make `wandb.log()` with different `step`.

catalyst/contrib/dl/runner/wandb.py
CrafterKolyan: I can change parameter name of the `epoch` to `step` and we can somehow (I don't know how and possibly this would be a hack) obtain `batch_idx` we can try to put it there OR there is a solution without this parameter:
Changing `wandb.log(metrics)` to `wandb.log(metrics, commit=False)`

We can also provide parameter `commit` for `_log_metrics` which would be default to `True`. Should we try this one solution?

CrafterKolyan: I've changed this one to `commit=False`. That's actually better solution.

Scitator:
@CrafterKolyan 
Don't you mind to merge with master branch and check, that everything is working correctly?
Moreover, due to loggers refactoring, would you like to contribute WandbLogger?

CrafterKolyan:
@Scitator Everything is working as expected.
image
There's actually new problem in `master` with `ResourceWarning` which I think is created by `# @TODO: add params for artefacts logging` commentaries. It's the same in `master`.
image

About `WandbLogger`: looking on `AlchemyLogger` I can say that `WandbLogger` seems absolutely the same with only extra artifacts logging. If you can do this copy paste job with your view of how `Logger`s should look like I would really appreciate it. I think I will waste too big amount of time on understanding how all of these `Logger`s work inside. If you provide some base `WandbLogger` which will look something like `AlchemyLogger` + `WandbRunner` I'll appreciate it. Also I am not sure how you actually should call these loggers and I really would like some **runnable minimal** example which shows **most functionality** of it (like in #684). Anyway maybe I'll do this new logger on my own but I am not sure I will find time. Thanks for the suggestion I will think about it in about 3 days.

Scitator:
Dear @CrafterKolyan ,
For an CustomLogger example. you can check out our [classification tutorial](https://github.com/catalyst-team/catalyst/blob/master/examples/notebooks/classification-tutorial.ipynb). I think, the implementation of WandbLogger would be straightforward without any problems.

CrafterKolyan:
@Scitator Ok, I l'll see that. But in my opinion `WandbLogger` should be in another change. So, can we accept this pull request?

Scitator:
@CrafterKolyan deal, looking forward to see your WandbLogger implementation.
Nevertheless, please fix the codestyle and deploy CI issues.

CrafterKolyan:
@Scitator Sorry, but I am not sure about which codestyle errors you are talking.
Running `make codestyle` I get `yapf` changing some files which are not connected to `wandb.py` at all. After adding them and running `make codestyle` again I get files changed back by `isort`. Am I doing something wrong? (I've installed everything with `pip install pre-commit` and `pip install -r requirements/requirements-dev.txt --no-cache-dir`).

`Flake8` gives only two codestyle errors:
```
catalyst/contrib/dl/runner/wandb.py:151:1: D102 Missing docstring in public method
catalyst/contrib/dl/runner/wandb.py:157:1: D101 Missing docstring in public class
```
I don't see any sense in fixing these ones because:
1. These runners are deprecated
2. These issues should have been there even before fix

About deploy CI issues:
As I see that's problems on your side, not on my. `master` has same issues.

CrafterKolyan:
Ok, I found out that these docstrings are pretty straight-forward and doesn't require any time to write them. Fixed.

Scitator:
@CrafterKolyan Could you please merge with master?
We have updated tests :)

CrafterKolyan:
@Scitator rebased

CrafterKolyan:
@Scitator I don't know why `Deploy` check starts `DL-GPU2`. Should I just recreate pull request in order for it to pass or can we just accept without this needless action?

PR #697, opened 2020-03-08
CheckpointCallback. Increase epoch on 1 when that loads checkpoint

lightforever:
### Description

Saved epoch was last we did. We should not repeat this epoch.

### Type of Change

## Checklist

- [ ] I have read the [Code of Conduct](https://github.com/catalyst-team/catalyst/blob/master/CODE_OF_CONDUCT.md) document.
- [ ] I have read the [Contributing](https://github.com/catalyst-team/catalyst/blob/master/CONTRIBUTING.md) guide.
- [ ] I have checked the code-style using `make check-codestyle`.

### DIFF
--- a/catalyst/core/callbacks/checkpoint.py
+++ b/catalyst/core/callbacks/checkpoint.py
@@ -35,8 +35,8 @@ def _load_checkpoint(*, filename, state: _State):
 
         if not state.stage_name.startswith("infer"):
             state.stage_name = checkpoint["stage_name"]
-            state.epoch = checkpoint["epoch"]
-            state.global_epoch = checkpoint["global_epoch"]
+            state.epoch = checkpoint["epoch"] + 1
+            state.global_epoch = checkpoint["global_epoch"] + 1
             # @TODO: should we also load,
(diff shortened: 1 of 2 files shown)


Scitator:
Not sure if, because we need to load the same state, as it was. Should we change the logic somethere else then?

lightforever:
@Scitator , done.

Description: The current behavior of CheckpointCallback is incorrect.

When stages>1, CheckpointCallback loads model weights every stage. This is caused due to the fact callbacks are being created each stage.

So, we need a static field in CheckpointCallback to store the fact that the resuming has happened.

That is the `resumed` field. About `if state.stage_name == stage_name:` - we should load while we don't catch an appropriate stage.

```
 if state.stage_name != stage:
            self._run_event("on_stage_end")
            return
```

If we loaded the next stage - we should stop the current one.

```
while state.epoch < state.num_epochs + 1:
```

if the epoch is the last one, this condition returns False, so we are going to the next stage without problems

Scitator:
It's only incorrect if it is shared between stages with same `resume` parameter. Nevertheless, passing CheckpointCallback directly to stage params would resolve this issue.

PR #716, opened 2020-03-21
initialize port by random value or value, which depends of SLURM_JOB_ID

AndreySheka:
## Description

DDP couldn't start twice at the same time on the same machine, because the master port is hardcoded.

## Type of Change

## Checklist

- [ ] I have read the [Code of Conduct](https://github.com/catalyst-team/catalyst/blob/master/CODE_OF_CONDUCT.md) document.
- [ ] I have read the [Contributing](https://github.com/catalyst-team/catalyst/blob/master/CONTRIBUTING.md) guide.
- [ ] I have checked the code-style using `make check-codestyle`.

### DIFF
--- a/catalyst/utils/distributed.py
+++ b/catalyst/utils/distributed.py
@@ -125,7 +125,7 @@ def get_slurm_params():
     current_node = os.environ["SLURMD_NODENAME"]
     master_node = socket.gethostbyname(nodes[0])
     cur_node_idx = nodes.index(current_node)
-    job_id = os.getenv("SLURM_JOB_ID")
+    job_id = os.environ["SLURM_JOB_ID"]
     master_port = str(5*10**4 + int(job_id) % 10**4)
     return cur_node_idx, num_nodes, master_node, master_port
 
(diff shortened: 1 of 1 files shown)


catalyst/utils/distributed.py
```diff
@@ -124,17 +125,20 @@ def get_slurm_params():
     current_node = os.environ["SLURMD_NODENAME"]
     master_node = socket.gethostbyname(nodes[0])
     cur_node_idx = nodes.index(current_node)
-    return cur_node_idx, num_nodes, master_node
+    job_id = os.getenv("SLURM_JOB_ID")
+    master_port = str(5*10**4 + int(job_id) % 10**4)
```
Scitator, line 129: could we first try to take `master_port` from os.env and randomly generate it only for None case?

AndreySheka: done

Scitator: not sure if :)

AndreySheka: We call this function if slurm available. By default, SLURM_JOB_ID must be defined. If SLURM_JOB_ID is None, it seems something wrong with slurm, and raising an exception is a good option.

PR #730, opened 2020-03-30
add `utils.get_network_output` as it used in catalyst.classification

bagxi:
## Description

[link](http://66.248.205.49:8111/viewLog.html?buildTypeId=Classification_Tests&buildId=24228&tab=buildLog&logTab=tree&filter=debug&expand=all&_focus=1800)

## Type of Change

## Checklist

- [x] I have checked the code style using `catalyst-check-codestyle` (`pip install -U catalyst-codestyle`).
- [x] I have written tests for all new methods and classes that I created.
- [x] I have read I need to click 'Login as guest' to see Teamcity build logs.

### DIFF
--- a/catalyst/utils/tests/test_torch.py
+++ b/catalyst/utils/tests/test_torch.py
@@ -1,0 +1,45 @@
+import torch
+from torch import nn
+
+from .. import torch as utils  # import
+
+
+def test_network_output():
+    """
+    Test for ``catalyst.utils.torch.get_network_output``
+    """
+    # case #0
+    net = nn.Identity()
+    assert utils.get_network_output(net, (1, 20)).shape == (1, 1, 20)
+
+    # case #1
+    net = nn.Linear(20, 10)
+    assert utils.get_network_output(net, (1, 20)).shape == (1, 1, 10)
+
+    # case #2
+    class Net(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.net1 = nn.Linear(20, 10)
+            self.net2 = nn.Linear(10, 5)
+
+        def forward(self, x, y):
+            z = torch.cat((self.net1(x), self.net2(y)), dim=-1)
+            return z
+
+    net = Net()
+    assert utils.get_network_output(net, (1, 20), (1, 10)).shape == (1, 1, 15)
+
+    # case #3
+    class Net(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.net = nn.Linear(20, 10)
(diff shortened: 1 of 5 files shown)


Scitator:
@bagxi could you please add some tests to https://github.com/catalyst-team/catalyst/tree/master/catalyst/utils/tests ? maybe, `tests_torch.py` would be good naming

for example, to test the feature into several cases,
1.
```
net = nn.Linear(20, 10)
input = (1, 20)
```
2.
```
class Net(nn.Module):
  net1 = ...
  net2 = ...

def forward(x, y):
  z = torch.cat(net1(x), net2(y))
  return z

net = Net()
input = ((1, 20), (1, 10))
```
3.
```
class Net(nn.Module):
  net1 = ...

def forward(some_input):
  z = net1(some_input)
  return z

net = Net()
input = {"some_input": (1, 20)}
```

catalyst/utils/tests/test_torch.py
```diff
@@ -0,0 +1,45 @@
...
+    # case #3
+    class Net(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.net = nn.Linear(20, 10)
+
+        def forward(self, input):
+            z = self.net(input["x"])
+            return z
+
```
Scitator: could we check here case like
```
def forward(self, *, x):
            z = self.net(x)
            return z
```
?

teamcity/py_codestyle.sh
```diff
@@ -10,4 +10,5 @@ pip install -r requirements/requirements-dev.txt
 pip install -r docs/requirements.txt
 
 catalyst-check-codestyle
-make check-docs
\ No newline at end of file
+pytest .
```
Scitator, line 13: could we also add test check to CONTRIBUTION.md?

Scitator (changes requested):

catalyst/utils/tests/test_torch.py
bagxi: I think no, as in this case we are going to test `**kwargs`, but the prototype of `torch.nn.Module.forward` is `forward(*input)` [pytorch docs](https://pytorch.org/docs/stable/nn.html#torch.nn.Module.forward)

Scitator: as4me, we should remove this functionality (`z = self.net(input["x"])`), 
or add support for `def forward(self, *, x):`
because, the case with `input["x"]` looks really hacky and complicated to understand

PR #816, opened 2020-05-19
fixed bugs in classification-tutorial

Podidiving:
## Before submitting

- [x] Did you read the [contribution guide](https://github.com/catalyst-team/catalyst/blob/master/CONTRIBUTING.md), Pull Request section?
- [ ] If you made a notable change (that affects users), did you update the [CHANGELOG](https://github.com/catalyst-team/catalyst/CHANGELOG.md)?

## Description
- After another release, `ImageReader` changed it's location from `catalyst.data` to `catalyst.contrib.data.cv`
- Also, it is now using `ToTensorV2` function, so albumentation must be updated on colab
- Runner now doesn't have `.callbacks`, they can be accessed via `experiment._callbacks`
- Fixed mistake in path to saved model

## Type of Change

### DIFF
--- a/examples/notebooks/classification-tutorial.ipynb
+++ b/examples/notebooks/classification-tutorial.ipynb
@@ -1208,7 +1208,7 @@
     "import numpy as np\n",
     "%matplotlib inline\n",
     "\n",
-    "class_counts = runner.experiment._callbacks[\"infer\"].class_counts\n",
+    "class_counts = runner.state._callbacks[\"infer\"].class_counts\n",
     "counts = [class_counts[x] for x in range(len(class_names))]\n",
     "\n",
     "plt.figure(figsize=(20, 9))\n",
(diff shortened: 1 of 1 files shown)


examples/notebooks/classification-tutorial.ipynb
```diff
@@ -46,7 +46,7 @@
    "outputs": [],
    "source": [
     "# for augmentations\n",
-    "!pip install albumentations\n",
+    "!pip install -U albumentations\n",
```
elephantmipt, line 49: Maybe we can just replace `!pip install catalyst`  with `!pip install catalyst[cv]` so we can control the version of albumentations by changing requirements_cv and prevent this example from failing. Also, in this case, we can remove `!pip install -U albumentations` as it is already in catalyst[cv] requirements. What do you think?

Podidiving: I have some doubts about your proposal:
1. catalyst itself has other dependencies we may want to use
2. we use new functionality of albumentations (ToTensorV2), so it's better to declare explicitly about installing new version

But maybe i'm wrong, idk

PR #842, opened 2020-06-10
Feature: wrapper for callbacks

ditwoo:
## Before submitting

- [x] Did you update the [CHANGELOG](https://github.com/catalyst-team/catalyst/CHANGELOG.md)?

## Type of Change

## PR review


catalyst/contrib/dl/callbacks/ignore_metric_callback.py
```diff
@@ -0,0 +1,78 @@
+from catalyst.core.callback import Callback, CallbackOrder
+from catalyst.core.runner import IRunner
+
+
+class IgnoreMetricCallback(Callback):
```
Scitator: just to be sure, could you please update docs with this new callbacks & changelog :)
I now it only draft yet :)

Scitator: speaking about architecture and implementation,
what do you think about `CallbackFilterCallback`?
not only with name filtering, but also with some condition one?
```python
callback_filter = CallbackFilterCallback(
  key={"train": "metric-1"}, 
  lambda={"valid": lambda key, value: isinstance(value, MetricCallback)},
  # or, even we can also add something like
  value={"valid": MetricCallback}
)
```

long story shot, with callback filtering, most of the time you have 2 options:
1. filter by name (`key` field)
2. filter by class (`value` field)
3. some very custom condition -> lambda

what do you think about such proposal?

Scitator: moreover, if we would like to think about filtering in a broader sense....
why do we filter only on loader name? during stage, we have 2 fors - epoch and loader one.
In this case, I think, we could filter not only on loader name, but also on epoch index.

nevertheless, still now sure about user friendly API for this :)

Scitator: maybe, only maybe, something like

```python
callback_filter = CallbackFilterCallback(
  condition=lambda epoch, loader: loader == "train",
  filter=lambda key, value: isinstance(value, MetricCallback),
)
```

ditwoo: @Scitator what do you think about `FilterCallback` or `SkipCallback` or `FilterMetricCallback` as a name for a class?
(I think that we can generate a better name without repeated *callback* word)

Also what do you think about use cases for arguments:

```python
callback = FilterCallback(
    loader0="to_filter",  # base case - ignore callback for loader
    loader1=["to_fitler1", "to_filter2"],  # ignore multiple callbacks
    loader2={"to_filter1": [1, 22, 333], "to_filter2": [1, 2, 3]},  # turn off callbacks for epochs
    loader3=lambda epoch, loader_name, loader_obj: loader_obj,  # filter with custom function
)
```

Scitator: @Ditwoo I think, `FilterCallback` would be good choice :)
First of all, we should not concentrate on Metric-based callbacks only.

catalyst/dl/experiment/config.py
```diff
@@ -430,6 +431,23 @@ def _get_callback(**params):
             return ConfigExperiment._get_callback(**wrapper_params)
         return callback
 
+    @staticmethod
+    def _is_same_types(first: Callback, second: Callback) -> bool:
```
Scitator: let's move this function to 
```
catalyst/dl/utils/callbacks.py
```
so it would be easier to use it from anywhere

Scitator: and what do you think about splitting it into two separate callbacks, like
```python
def get_original_callback(callback) -> callback:
  while not callback and is-wrapper:
    return callback.callback  # unwrap

def check_isinstance(first, second) -> bool:
  first = get_original_callback(first)
  second = get_original_callback(second)
  output = isinstance(first, second)
  return output
```

bin/tests/check_dl_core_wrapper_callback.sh
```diff
@@ -0,0 +1,265 @@
...
+LOG_MSG='pipeline 00'
+echo ${LOG_MSG}
+
+LOGDIR=./tests/logs/_tests_dl_callbacks
+CHECKPOINTS=${LOGDIR}/checkpoints
+LOGFILE=${CHECKPOINTS}/_metrics.json
+EXP_OUTPUT=./tests/output.txt
+
+PYTHONPATH=./examples:./catalyst:${PYTHONPATH} \
+  python3 -c "
```
Scitator, line 72: btw, could you please move your tests to .py files,
like https://github.com/catalyst-team/catalyst/blob/master/catalyst/contrib/dl/callbacks/tests/test_tracer_callback.py
?

catalyst/contrib/dl/callbacks/periodic_loader_callback.py
```diff
@@ -12,13 +12,31 @@ class PeriodicLoaderCallback(Callback):
...
+    3 epochs and ``valid_additional`` - every 5 epochs:
+
+    .. code-block:: python
+
+        from catalyst.dl import (
+            SupervisedRunner, PeriodicLoaderRunnerCallback,
+        )
+        runner = SupervisedRunner()
+        runner.train(
+            ...
```
Scitator, line 27: would you like to add `loaders={"train_additional": ..., "valid": ..., "valid_additional":...}`?

catalyst/core/callbacks/wrapper.py
```diff
@@ -0,0 +1,298 @@
...
+    need to wrap callback:
+
+    .. code-block:: python
+
+        from catalyst.dl import (
+            SupervisedRunner, WrapperCallback, CriterionCallback
+        )
+        runner = SupervisedRunner()
+        runner.train(
+            ...
```
Scitator: 🤔  what do you think about full-featured examples?
we have really tiny one:
```python
import torch
from torch.utils.data import DataLoader, TensorDataset
from catalyst.dl import SupervisedRunner

num_samples, num_features = int(1e4), int(1e1)
X, y = torch.rand(num_samples, num_features), torch.rand(num_samples)
loader = DataLoader(TensorDataset(X, y), batch_size=32, num_workers=1)
loaders = {"train": loader, "valid": loader}

model = torch.nn.Linear(num_features, 1)
criterion = torch.nn.MSELoss()
optimizer = torch.optim.Adam(model.parameters())
scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer, [3, 6])

runner = SupervisedRunner()
runner.train(
    model=model, criterion=criterion, optimizer=optimizer, scheduler=scheduler,
    loaders=loaders, logdir="./logdir", num_epochs=8, verbose=True,
)
```

catalyst/core/callbacks/wrapper.py
```diff
@@ -0,0 +1,298 @@
...
+            callback: CriterionCallback
+          ...
+
+    """
+
+    def __init__(
+        self,
+        base_callback: Callback,
+        loaders: LOADERS = None,
+        filter_fn: Union[str, FILTER_FN] = None,
```
Scitator, line 229: would you like to add `epoch` support to run callback every `epoch` epochs? 😂

catalyst/dl/utils/callbacks.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Returns:
+        callback inside wrapper
+    """
+    while isinstance(callback, WrapperCallback):
+        callback = callback.callback
+    return callback
+
+
+def check_isinstance(first: Callback, second: Callback) -> bool:
```
Scitator: `check_callback_isinstance`?

docs/api/core.rst
```diff
@@ -129,6 +129,12 @@ Validation
     :undoc-members:
     :show-inheritance:
 
+Wrapper
+~~~~~~~~~~~~~~~~~~~~~~
+.. automodule:: catalyst.core.callbacks.wrapper
```
Scitator, line 134: ❤️

catalyst/contrib/dl/callbacks/periodic_loader_callback.py
ditwoo: yep, i agree, need to add this

catalyst/core/callbacks/wrapper.py
ditwoo: sure

catalyst/core/callbacks/wrapper.py
```diff
@@ -0,0 +1,361 @@
...
+            optimizer=optimizer,
+            scheduler=scheduler,
+            loaders=loaders,
+            logdir="./logdir",
+            num_epochs=8,
+            verbose=True,
+            callbacks=[
+                WrapperCallback(
+                    base_callback=CriterionCallback(),
+                    loaders=["valid"]
```
Scitator: 🤔  based on provided example, should we split it's functionality to 
```
loaders=[] # loader names, where callback should be executed
ignore_loaders=[] # loader names, where we should ignore the callback
```
?

just wondering, how it would be more user-friendly

Scitator: btw, it also looks like this examples should not work because by default our metric - "loss" on "valid" loader, that is not computed here :)

Scitator: I think, we could easily use `dl.AccuracyCallback` instead

catalyst/core/callbacks/wrapper.py
```diff
@@ -289,6 +354,9 @@ def on_loader_start(self, runner: IRunner) -> None:
         if self.filter_fn is not None:
             self._is_active = not self.filter_fn(stage, epoch, loader)
 
+        # import inspect
+        # print(">>>", inspect.getsource(self.filter_fn))
```
Scitator: not sure if we need this :)

catalyst/core/callbacks/wrapper.py
```diff
@@ -29,18 +31,28 @@ def _filter_fn_from_epochs(
 
     if isinstance(epochs, (list, tuple)):
         epochs = sorted(set(epochs))
-        filter_fn = lambda stage, epoch, loader: epoch in epochs
+        if reverse_condition:
+            filter_fn = lambda stage, epoch, loader: epoch not in epochs
+        else:
+            filter_fn = lambda stage, epoch, loader: epoch in epochs
```
Scitator, line 40: do you think epoch list would be user friendly?
from my point of view most of the time you want to execute something `one time per N epochs`

maybe we should change the above condition to `epoch % specified_epoch == 0`?

catalyst/core/callbacks/wrapper.py
ditwoo: it was for debug 😃

catalyst/core/callbacks/wrapper.py
ditwoo: @Scitator not everywhere.
i think we should keep both:
- if specified number (for example `N`) then disable/enable callback every `N` epochs
- if specified list of numbers then disable/enable callback at specified epochs

also, if we touch `epochs` did we expect similar logic from `loaders` argument?
in current implementation it supports only list of epochs.
do we need the same logic for `loaders`:

```
loaders={
     <loader 1>: <number>,  # should we do some periodical action for this loader?
     <loader 2>: [<number 1>, <number 2>, <number 3>],  # disable/enable callback for specified epochs
}
```

?

Scitator: 🤔  I think, you are right about 2 cases support for epochs
nevertheless, for loaders I suppose we only have enable/disable action

catalyst/core/callbacks/wrapper.py
```diff
@@ -8,6 +8,113 @@
 FILTER_FN = Callable[[str, int, str], bool]
 
 
+class WrapperCallback(Callback):
```
Scitator: maybe,  we should move it to `catalyst/core/callback`?

ditwoo: sure

Scitator (approved):

PR #873, opened 2020-07-03
fix empty loader

AlekseySh:
We don't check if DataLoader is empty before the training start, that is why can produce confused error message like:
`AssertionError: accuracy value is not available by the epoch end`

I gonna fix it.


AlekseySh:
@Scitator 
Check my implementation, please.

A also have a question: 
- Should I use my own `EmptyLoaderException` instead of `AssertionError` to catch it in tests more directly? If yes, where should I put such an exception?

```
def test_cathing_empty_loader() -> None:
    try:
        run_train_with_empty_loader()
    except AssertionError:  # <------- here
        assert True
```

catalyst/core/runner.py
```diff
@@ -826,6 +826,8 @@ def _run_loader(self, loader: DataLoader) -> None:
         Args:
             loader (DataLoader): dataloader to iterate
         """
+        assert loader, f"DataLoader with name {self.loader_name} is empty."
```
Scitator: what do you think about copying this assert to `run_epoch` too?
It looks easier to check them instantly on epoch start, rather than wait until (maybe some empty) loader run.

Scitator: nevertheless, this assert is valid too, because we could change the loaders with the callbacks.
btw, why non't you like to use `len(loader) > 0)`? I looks much easier to understand, isn't it?

AlekseySh: I think that `_run_loader` is better because we can change the loaders via callbacks, as you said.

As for `len(loader) > 0` -- ok, I will fix it.

AlekseySh:
I also replaced Assert with Exception
@Scitator pls check

catalyst/core/runner.py
Scitator: Yup, we definitely need that check on `_run_loader`, but why should not we also add some kind of pre-check on `_run_epoch`? what do you think about it?

CHANGELOG.md
```diff
@@ -64,7 +64,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Fixed
 
 - added missed dashes in docker perfixes ([#828](https://github.com/catalyst-team/catalyst/issues/828))
-
+- handle empty loader in Runner
```
Scitator: look like we need PR link here :)

AlekseySh: done

catalyst/core/runner.py
AlekseySh: Okay, done

PR #893, opened 2020-07-22
Feature: TPU support

ditwoo:
## Before submitting

## Type of Change

## PR review


catalyst/core/callbacks/optimizer.py
```diff
@@ -70,14 +83,20 @@ def grad_step(
...
+                function, default is ``None``
         """
         for group, wd in zip(optimizer.param_groups, optimizer_wds):
             if wd > 0:
                 for param in group["params"]:
                     param.data = param.data.add(-wd * group["lr"], param.data)
             if grad_clip_fn is not None:
                 grad_clip_fn(group["params"])
-        optimizer.step()
+        if xla_args is not None:
```
Scitator: let's define `callback._optimizer_step` and redefine it based on xla availability :)

catalyst/utils/components.py
```diff
@@ -65,6 +65,15 @@ def process_components(
 
     model: Model = maybe_recursive_call(model, "to", device=device)
 
+    if optimizer is not None:
+        if isinstance(optimizer, dict):
+            optimizer = {
+                k: type(opt)(model[k], **opt.defaults)
```
Scitator: `model[k].parameters()`?

catalyst/utils/torch.py
```diff
@@ -15,6 +15,13 @@
 from catalyst.tools.typing import Device, Model, Optimizer
 from catalyst.utils.dict import merge_dicts
 
+try:
+    import torch_xla.core.xla_model as xm
+
+    XLA_DEVICE = xm.xla_device()
```
Scitator: could we move `IS_XLA_AVAILABLE` to `catalyst/tools/settings`?
@bagxi what do you tnink?

catalyst/core/callbacks/checkpoint.py
```diff
@@ -199,6 +199,11 @@ def on_exception(self, runner: IRunner):
         if not utils.is_exception(exception):
             return
 
+        if runner.device.type == "xla":
```
Scitator, line 202: could we define it as global `MODEL_SAVE_FN`?

ditwoo: What do you think about `save_fn` as runner function?

Scitator: no... generally speaking I would like to move all imports to the top on the file
like
```
from catalyst.setting import IS_XLA_AVAILABLE
if IS_XLA_AVAILABLE:
   xla_model_save = ...
else:
   ...
torch_model_save = ...
```
and here:

saver_fn = xla_model_save if IS_XLA_AVAILABLE and runner.type == "xla" else torch_model_save

Scitator: btw, I also think, that we could pass this setting to BaseCheckpointClass, rather than duplicate this functionality in both of them

catalyst/core/callbacks/optimizer.py
```diff
@@ -10,6 +11,13 @@
 
 logger = logging.getLogger(__name__)
 
+try:
+    from torch_xla.core.xla_model import optimizer_step as xla_optimizer_step
+except ModuleNotFoundError:
+
+    def xla_optimizer_step(optimizer, *args, **kwargs):
```
Scitator: why would you override it?

catalyst/core/callbacks/optimizer.py
```diff
@@ -70,14 +85,16 @@ def grad_step(
             optimizer_wds (List[float]): list of weight decay parameters
                 for each param group
             grad_clip_fn (Callable): function for gradient clipping
+            optmizert_step_fn (Callable): optimizer step function
```
Scitator: now it looks too complicated, let's just remove static and use self.optimizer_step

catalyst/core/callbacks/optimizer.py
```diff
@@ -88,6 +105,16 @@ def on_stage_start(self, runner: IRunner) -> None:
...
         )
+        if runner.device.type == "xla":
+            if self.use_xla_barrier:
+                self._optimizer_step_fn = lambda optimizer: xla_optimizer_step(
+                    optimizer, barrier=True
+                )
+            else:
+                self._optimizer_step_fn = xla_optimizer_step
+        else:
+            self._optimizer_step_fn = lambda optimizer: optimizer.step()
```
Scitator: and I think it's easier to move this logic to `def _optimizer_step(self, ...)`

catalyst/core/callbacks/optimizer.py
```diff
@@ -168,10 +195,14 @@ def on_batch_end(self, runner: IRunner) -> None:
             loss.backward()
 
         if need_gradient_step:
+            xla_args = (
+                {"barrier": self.use_xla_barrier} if self.is_xla else None
+            )
```
Scitator: why do you need it here?

catalyst/utils/checkpoint.py
```diff
@@ -1,6 +1,6 @@
 # flake8: noqa
 # @TODO: code formatting issue for 20.07 release
```
Scitator: so, looks like the time has come to remove this barge :)

catalyst/utils/checkpoint.py
```diff
@@ -84,11 +84,27 @@ def save_checkpoint(
     is_best: bool = False,
     is_last: bool = False,
     special_suffix: str = "",
+    saver_fn: Callable = torch.save,
```
Scitator, line 96: just `save_fn`? btw, what do you think about simple `is_xla_model` flag? so
```
if is_xla_model:
   xla_model_save(checkpoint, filename)
else:
   torch.save(checkpoint, filename)
```
?

catalyst/utils/components.py
```diff
@@ -65,6 +65,17 @@ def process_components(
...
+    if optimizer is not None:
+        if isinstance(optimizer, dict):
+            optimizer = {
+                k: type(opt)(model[k].parameters(), **opt.defaults)
+                for k, opt in optimizer.items()
+            }
+        else:
+            optimizer = type(optimizer)(
+                model.parameters(), **optimizer.defaults
+            )
```
Scitator: really tricky solution :)
I understand why, but... are you sure, that opt.defaults have all required parameters?

ditwoo: I think it will work in most cases.
I've checked optimizers in pytorch and saw that all arguments wraps in defaults dict and used in initialization of super class [Optimizer](https://pytorch.org/docs/stable/optim.html#torch.optim.Optimizer)

But this approach is bad because this implementation don't know anything about user defined learning rates for layers and we need to somehow handle this case.

catalyst/utils/torch.py
bagxi: We can do so. But keep in mind that settings responsible only for imports (e.g. kornia) for now

catalyst/utils/components.py
```diff
@@ -47,13 +48,24 @@ def process_components(
...
     distributed_params = distributed_params or {}
     distributed_params = copy.deepcopy(distributed_params)
     distributed_params.update(get_distributed_params())
 
+    if device is None and IS_XLA_AVAILABLE:
+        raise ValueError(
+            "TPU device is available. "
+            "Please move model, optimizer and scheduler (if present) "
+            "to TPU device manualy and specify a device or "
+            "use CPU device."
```
Scitator, line 66: could you please specify an example, like 
```
"TPU device is available. "
            "Please move model, optimizer and scheduler (if present) "
            "to TPU device manualy and specify a device for runner with `Runner(device={your_tpu_device_here})`"
```
what do you think about it?

catalyst/utils/checkpoint.py
```diff
@@ -84,20 +93,42 @@ def save_checkpoint(
...
```
Scitator, line 96: maybe, just `save_fn`?

catalyst/core/callbacks/optimizer.py
```diff
@@ -33,6 +39,8 @@ def __init__(
             grad_clip_params (dict): params for gradient clipping
             decouple_weight_decay (bool): If True - decouple weight decay
                 regularization.
+            use_xla_barrier (bool): use barrier option if used xla,
```
Scitator: could you please add link-to-docs for xla barrier?

catalyst/core/callbacks/optimizer.py
```diff
@@ -88,6 +100,16 @@ def on_stage_start(self, runner: IRunner) -> None:
...
         )
+        if runner.device.type == "xla":
+            if self.use_xla_barrier:
+                self._optimizer_step_fn = lambda optimizer: optimizer_step(
+                    optimizer, barrier=True
+                )
+            else:
+                self._optimizer_step_fn = optimizer_step
+        else:
+            self._optimizer_step_fn = lambda optimizer: optimizer.step()
```
Scitator: could we just specify 2 extra methods for callback?
```python
def _optimizer_step_gpu()...
  self._optimizer.step()

def _optimizer_step_tpu()
  ???
  
```

Scitator: just don't like the lambda usage :)

catalyst/core/callbacks/checkpoint.py
```diff
@@ -335,6 +341,7 @@ def __init__(
         self.metrics_history = []
 
         self._keys_from_state = ["resume", "resume_dir"]
+        self._saver_fn: Callable = None
```
Scitator: `save_fn`?

Scitator: `checkpoint_save_fn`?

catalyst/core/callbacks/optimizer.py
ditwoo: sure, thanks

PR #894, opened 2020-07-23
Pruning

elephantmipt:
## Before submitting

- [ ] Did you read the [contribution guide](https://github.com/catalyst-team/catalyst/blob/master/CONTRIBUTING.md)?
- [ ] Did you check the code style? `catalyst-make-codestyle && catalyst-check-codestyle` (`pip install -U catalyst-codestyle`).

## Type of Change

## PR review


catalyst/core/callbacks/__init__.py
```diff
@@ -5,6 +5,11 @@
     CheckpointCallback,
     IterationCheckpointCallback,
 )
+
+import torch
```
Scitator: could you please move it to the top?

catalyst/core/callbacks/__init__.py
```diff
@@ -5,6 +5,11 @@
...
+
+if torch.__version__ > "1.4":
```
Scitator: btw, let's move this check to `catalyst/tools/settings.py` for convenience

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,144 @@
...
+                if self._to_be_pruned(name):
+                    if isinstance(self.key_to_prune, str):
+                        prune.remove(module, self.key_to_prune)
+                    elif isinstance(self.key_to_prune, list):
+                        for key in self.key_to_prune:
+                            prune.remove(module, key)
+            except ValueError:
+                pass
+
+    def on_epoch_end(self, runner) -> None:
```
Scitator: could you please add types here?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,144 @@
...
+        On epoch end action.
+
+        Active if prune_on_epoch_end is True.
+        Args:
+            runner: runner for your experiment
+        """
+        if self.prune_on_epoch_end and runner.num_epochs != runner.epoch:
+            self._run_pruning(runner)
+
+    def on_stage_end(self, runner) -> None:
```
Scitator: could you please add types here?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def _remove_reparametrization(self, runner: "IRunner"):
+        for name, module in runner.model.named_modules():
+            try:
+                if self._to_be_pruned(name):
+                    if isinstance(self.key_to_prune, str):
+                        prune.remove(module, self.key_to_prune)
+                    elif isinstance(self.key_to_prune, list):
+                        for key in self.key_to_prune:
+                            prune.remove(module, key)
```
Scitator: could we just make `self.key_to_prune` list by design?
and rename to `self.keys_to_prune`?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,144 @@
...
+        elif isinstance(self.key_to_prune, list):
+            for key in self.key_to_prune:
+                self.pruner_fn(module, name=key, amount=self.amount)
+
+    def _to_be_pruned(self, layer_name):
+        if self.layers_to_prune is None:
+            return True
+        return layer_name in self.layers_to_prune
+
+    def _run_pruning(self, runner: "IRunner"):
```
Scitator: do you think we need full runner here? maybe `model` would be enough?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,144 @@
...
+                "Model won't be pruned by this callback."
+            )
+        self.remove_reparametrization = remove_reparametrization
+        self.key_to_prune = key_to_prune
+        self.amount = amount
+        self.reinitialize_after_pruning = reinitialize_after_pruning
+        self.layers_to_prune = layers_to_prune
+
+    @staticmethod
+    def _weight_reset(m):
```
Scitator: could we move this function to `catalyst/utils/initialization.py`?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+    Pruning Callback
+
+    This callback is designed to prune network parameters
+    during and/or after training.
+    """
+
+    def __init__(
+        self,
+        pruner_fn: Callable,
```
Scitator: let's parametrise as string and use like
```python
import torch.nn.utils.prune as pruner_fns
prune_name = "l1_unstructured"
prune_fn =  pruner_fns.__dict__[prune_name]
```

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,144 @@
...
+        if isinstance(self.key_to_prune, str):
+            self.pruner_fn(module, name=self.key_to_prune, amount=self.amount)
+        elif isinstance(self.key_to_prune, list):
+            for key in self.key_to_prune:
+                self.pruner_fn(module, name=key, amount=self.amount)
+
+    def _to_be_pruned(self, layer_name):
+        if self.layers_to_prune is None:
+            return True
+        return layer_name in self.layers_to_prune
```
Scitator: could we make it `self.layers_to_prune is None or layer_name in self.layers_to_prune`?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,186 @@
...
+
+from catalyst.core import Callback, CallbackOrder, IRunner
+from catalyst.utils.initialization import weight_reset
+
+PRUNING_FN = {
+    "l1_unstructured": prune.l1_unstructured,
+    "random_unstructured": prune.random_unstructured,
+    "ln_structured": prune.ln_structured,
+    "random_structured": prune.random_structured,
+}
```
Scitator: you can just use prune.__dict__[{fn_name_here}]

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,186 @@
...
+                    raise e
+
+        if pruned_modules == 0:
+            raise Exception(
+                f"There is no {self.keys_to_prune} key in your model"
+            )
+        if self.reinitialize_after_pruning:
+            model.apply(weight_reset)
+
+    def _remove_reparametrization(self, runner: "IRunner"):
```
Scitator: do you need runner here?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,186 @@
...
+
+    @staticmethod
+    def _wrap_pruning_fn(pruning_fn, *args, **kwargs):
+        return lambda module, name, amount: pruning_fn(
+            module, name, amount, *args, **kwargs
+        )
+
+    def _prune_module(self, module):
+        for key in self.keys_to_prune:
+            self.pruning_fn(module, name=key, amount=self.amount)
```
Scitator: do we really need this method? could we just use the code in `_run_pruning`?

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,186 @@
...
+            except AttributeError as e:
+                if self.layers_to_prune is not None:
+                    raise e
+
+        if pruned_modules == 0:
+            raise Exception(
+                f"There is no {self.keys_to_prune} key in your model"
+            )
+        if self.reinitialize_after_pruning:
+            model.apply(weight_reset)
```
Scitator: could we make this function static, or just move it to catalyst.utils?
for example, make it:
```python
def prune_model(model, layers_to_prune: optional[str] = None, reinitialize_after_pruning=False, keys_to_prune, amount): 
   ...
```
?

catalyst/core/callbacks/pruning.py
elephantmipt: `torch.nn.utils.prune` module also has `global` pruning type methods, which has different APIs. I don't think that putting different behaviour in one callback is a good idea, despite the fact that basically they do the same thing.
Here are examples:
`prune.random_unstructured(module, name="weight", amount=0.3)`
`prune.ln_structured(module, name="weight", amount=0.5, n=2, dim=0)  # we also need dim`
There is almost no difference.
And here is example with `global` pruning:
```python
parameters_to_prune = (
    (model.conv1, 'weight'),
    (model.conv2, 'weight'),
    (model.fc1, 'weight'),
    (model.fc2, 'weight'),
    (model.fc3, 'weight'),
)

prune.global_unstructured(
    parameters_to_prune,
    pruning_method=prune.L1Unstructured,
    amount=0.2,
)
```
So the question is: Do we need `GlobalPruningCallback` or I should integrate global pruning methods to existing callback?

elephantmipt: Also as you can see global pruning method is a kind of more high-level as it is using other pruning_fn

catalyst/core/callbacks/pruning.py
```diff
@@ -0,0 +1,157 @@
...
+                "Model won't be pruned by this callback."
+            )
+        self.remove_reparametrization = remove_reparametrization
+        self.keys_to_prune = keys_to_prune
+        self.amount = amount
+        self.reinitialize_after_pruning = reinitialize_after_pruning
+        self.layers_to_prune = layers_to_prune
+
+    @staticmethod
+    def _wrap_pruning_fn(pruning_fn, *args, **kwargs):
```
Scitator: let's make it a function, just before the callback

catalyst/tools/settings.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         use_libjpeg_turbo: bool = False,
         nmslib_required: Optional[bool] = None,
         transformers_required: Optional[bool] = None,
+        pytorch_14: Optional[bool] = None,
```
Scitator: `is_pytorch_above_v14`?

catalyst/utils/initialization.py
```diff
@@ -75,4 +75,17 @@ def outer_init(layer: nn.Module) -> None:
             nn.init.uniform_(layer.bias.data, -v, v)
 
 
-__all__ = ["get_optimal_inner_init", "outer_init"]
+def weight_reset(module: nn.Module):
```
Scitator, line 78: `reset_weights`? btw, do we really need this method? could we just use `module.reset_parameters()`

catalyst/utils/pruning.py
```diff
@@ -0,0 +1,51 @@
...
+def prune_model(
+    model: Module,
+    pruning_fn: Callable,
+    keys_to_prune: List[str],
+    amount: Union[float, int],
+    layers_to_prune: Optional[List[str]] = None,
+    reinitialize_after_pruning: Optional[bool] = False,
+) -> None:
+    """
+    @TODO
```
Scitator: docs please 🙏

Scitator (commented):
btw, what do you think about pruning integration to Config API?
for example, 
```
catalyst-dl prune --logdir=/path/to/logdir --some-other-params
```
?

catalyst/utils/initialization.py
elephantmipt: in this case we need to create loop like
```python
for _name, module in model.named_parameters():
    try:
        module.reset_parameters()
    except AttributeError as ex:
        pass
```
Instead of `model.apply(weight_reset)`
I think the second variant is more preferable.

catalyst/core/callbacks/pruning.py
Scitator: 🤔  looks like good option for todo

catalyst/tools/settings.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         use_libjpeg_turbo: bool = False,
         nmslib_required: Optional[bool] = None,
         transformers_required: Optional[bool] = None,
+        is_pytorch_above_v14: Optional[bool] = None,
```
Scitator: not sure if this is correct, in our case with torch version this should work like `IS_XLA_AVAILABLE`, isn't it?
you just cann't turn on/off torch version :)

catalyst/utils/initialization.py
```diff
@@ -75,4 +75,17 @@ def outer_init(layer: nn.Module) -> None:
...
+    """
+    Reseting module parameters.
```
Scitator: `...if possible`?

Issue #912, opened 2020-08-10
Hydra/OmegaConf configuration structure


otherman16:
## 🚀 Feature Request

[Hydra](https://hydra.cc) provides interesting [features](https://hydra.cc/docs/intro) to create hierarchical configuration by composition and override it through config files and the command line.
I think that this ability will be useful for catalyst pipelines with configuration file.
Now one configuration file contains all pipeline settings inside, but if we split it to for example model_params/data_params/runner_params/and so on then we will be able to get configuration structure distributed in sence. This will allow us to quickly run experiments with different combinations of settings and also change settings in the command line

Example pipeline:
```bash
pipeline/
    conf/
        config.yaml
        model/
            resnet.yaml
            mobilenet.yaml
            efficientnet.yaml
        data/
            imagenet.yaml
            coco.yaml
        runner/
            runner1.yaml
            runner2.yaml
    src/
        callbacks.py
        model.py
        ...
```
Run example:
```catalyst-dl run config=conf/config.yaml model=resnet data=imagenet runner=runner1```

### Possible solution
1. Create class HConfigExperiment class
2. Change catalyst-dl run script

### Motivation

Combining configuration parts for different experiments. Change settings in command line.

### Additional context

I've already written simple MNIST classification pipeline with hydra configuration structure. But I can run it only with ```python train.py``` command.

**Comment by omry** [comment]
Hi,
Definitely take a look at [Structured Configs](https://hydra.cc/docs/next/tutorials/structured_config/intro) in Hydra 1.0.
It's taking Hydra to the next level, and is especially useful for frameworks.

**Comment by Scitator** [comment]
@otherman16 Hi, looks interesting, but how does it differ from current multi-config behaviour?
```bash
catalyst-dl run config conf/config-main.yaml conf/config-model-resnet.yaml conf/config-extra-new-name.yaml
```

otherman16:
@Scitator They are almost the same but with Hydra/OmegaConf you can also:
1. Change specific settings in command line. Example:
```catalyst-dl run config=conf/config.yaml model=resnet data=imagenet runner=runner1 model.resnet=50/101/...```
2. Command line [Tab Completion](https://hydra.cc/docs/next/tutorials/basic/running_your_app/tab_completion)
3. Select default configuration
[...]

PR #943, opened 2020-09-23
OmegaConf support for "catalyst-dl run"

otherman16:
## Before submitting

## Description

OmegaConf support for `catalyst-dl run ...`

## Related Issue

[Hydra/OmegaConf configuration structure #912](https://github.com/catalyst-team/catalyst/issues/912)

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)

Breaking change: OmegaConf command line arguments style:
`--path=test.yaml:str` -> `args.path=test.yaml`
`--stage/foo=bar:str` -> `stage.foo=bar`

## PR review


catalyst/utils/config.py
```diff
@@ -5,11 +5,36 @@
 from pathlib import Path
 import re
 
+from omegaconf import DictConfig, ListConfig
 import yaml
 
 LOG = getLogger(__name__)
 
 
+def omegaconf_to_dict(config):
```
omry: Use the supported `OmegaConf.to_container(cfg, resolve=True)` to convert an OmegaConf object to a primitive container (dict or list).
You do not need to implement this function (I am also guessing that this implementation will fail in various cases covered by the official implementation).

catalyst/utils/tests/test_config.py
```diff
@@ -17,30 +17,29 @@ def test_parse_config_args():
...
-    assert "zero" in configuration["stages"]
-    assert configuration["stages"]["zero"] == "cero"
-    assert configuration.get("args") is not None
-    assert configuration["args"]["path"] == "test.yml"
-    assert configuration["args"]["C"] == "like"
-    assert configuration["args"]["command"] == "run"
+    assert "logdir" in configuration["args"]
+    assert configuration["args"]["logdir"] == "logdir"
+    assert "autoresume" in configuration["args"]
+    assert configuration["args"]["autoresume"] == "last"
```
omry: If your config is now OmegaConf config, you can switch to a nicer attribute style access, e.g `configuration.args.logdir`.

catalyst/utils/config.py
otherman16: Cool! I couldn't find this method in docs. Very thanks!

catalyst/utils/tests/test_config.py
otherman16: Ok. But this feature will be in next iteration I think.

catalyst/utils/config.py
omry: [Here](https://omegaconf.readthedocs.io/en/latest/usage.html#omegaconf-to-container).

catalyst/dl/scripts/run.py
```diff
@@ -105,18 +107,17 @@ def main_worker(args, unknown_args):
     utils.set_global_seed(args.seed)
     utils.prepare_cudnn(args.deterministic, args.benchmark)
 
-    config.setdefault("distributed_params", {})
+    config = OmegaConf.to_container(config, resolve=True)
```
omry: This brings me to the next question:
What happens if you use DictConfig as is?
for most purposes, it's a drop in replacement to a regular dict.

Scitator (dismissed):
For OmegaConf we need to save backward compatibility and only add an option to use hydra-like configs for user.

catalyst/dl/scripts/run.py
otherman16: I have already try it, but in some cases Dict can not be replaced by DictConfig (for example json.dump() method).

catalyst/utils/parser.py
```diff
@@ -3,13 +3,14 @@
...
-from catalyst.utils.config import load_config
-from catalyst.utils.dict import merge_dicts
+from omegaconf import OmegaConf
 
 
 def parse_config_args(*, config, args, unknown_args):
     """@TODO: Docs. Contribution is welcome."""
     for arg in unknown_args:
+        if not arg.startswith("-"):
+            continue
```
Scitator, line 13: why do we need this?

requirements/requirements.txt
```diff
@@ -21,3 +21,6 @@ matplotlib
 pandas>=0.22
 scikit-learn>=0.20
 GitPython>=3.1.1
+
+# OmegaConf
+omegaconf
```
Scitator, line 26: is it possible to make it optional dependency?

catalyst/utils/parser.py
```diff
@@ -94,6 +96,10 @@ def parse_args_uargs(args, unknown_args):
                 arg_value = value
             setattr(args_copy, key, arg_value)
 
+    print(config.pretty())
```
Scitator, line 99: not sure about this :)

catalyst/utils/parser.py
```diff
@@ -74,10 +75,11 @@ def parse_args_uargs(args, unknown_args):
...
 
     # load params
-    config = {}
+    configs = []
     for config_path in args_copy.configs:
-        config_part = load_config(config_path, ordered=True)
-        config = merge_dicts(config, config_part)
+        configs.append(OmegaConf.load(config_path))
+    cli_config = OmegaConf.from_cli()
+    config = OmegaConf.merge(*configs, cli_config)
```
Scitator, line 82: could you please provide an example of usage for the new functionality?

catalyst/utils/parser.py
otherman16: "+"

catalyst/utils/parser.py
otherman16: Example: `catalyst--dl run --foo=bar:str foo=bar`
`--foo=bar:str` - catalyst-style argument
`foo=bar` - omegaconf-style argument
Both arguments will be processed in this block, but "foo=bar" cannot be parsed with
`value_content, value_type = value.rsplit(":", 1)` line because "foo=bar".rsplit(":", 1) gives 1 value instead 2.
So I have added this condition to separate catalyst-style arguments and omegaconf-style arguments.

requirements/requirements.txt
otherman16: No. In current realisation all confguration "*.yaml" files are parsed and merged with OmegaConf.
If we want to leave catalyst-config functionality as it is and provide new functionality independently and make it optional I think we should create independent scripts `omegaconf-run.py` and `omegaconf-parser.py` for example.

catalyst/utils/parser.py
otherman16: `catalyst-dl run --configs=config1.yaml,config2.yaml args.expdir=/path/to/src --logdir=/path/to/logs stages.data_params.batch_size=32 stages.train.state_params.num_epochs=40 --stages/train/callback_params/saver/save_n_best=2:int`
where can I write this?

requirements/requirements.txt
omry: I think if you create alternative entry points you should use Hydra for it and not invent your own mechanism.

otherman16: Agree. Hydra encapsulates configuration parsing. It's convenient.

omry: @Scitator,
The original logic with merge_dicts below shows that catalyst is already solving a subset of the problems Hydra is solving (It is a form of config composition).
Large ML frameworks are switching to Hydra for it's config composition capabilities (recent example is fairseq).
Doing it in a way that retains backward compatibility is hard but possible, at least judging based on their experience.

I think embracing Hydra is a strategic project decision. In the short term using OmegaConf as the underlying config abstraction will be beneficial and will make it easier to interface with Hydra if and when the decision happens.

Some of the benefits:
* Across the codebase, you can use `cfg.foo.bar` instead of `cfg["foo"]["bar"]`
* Usage of interpolation (think of them as pointers in the config).
* Rudimentary command line integration (Hydra has a similar but more powerful CLI).
* More robust config merging capabilities
* Type safe configs via dataclasses (see Structured Configs in the OmegaConf docs).

catalyst/utils/parser.py
Scitator: and what about more complex case like
```
catalyst--dl run --/stages/stage1/optimizer_params/opmizer=MyCustomOptimizer:str
```
or
```
catalyst--dl run --/stages/stage1/optimizer_params/opmizer_staps='[10, 20, 30]':list
```

is it possible with omega-conf? @otherman16 @omry

Scitator: just wondering how omega-conf could resolve different python types

catalyst/utils/parser.py
Scitator: I think you could add this to our pipelines test cases...
we have a bunch of them 😄 
https://github.com/catalyst-team/catalyst/blob/master/bin/tests/check_dl_cv.sh

requirements/requirements.txt
Scitator: firstly, I don't think we need new entry point, we could just add a new argument, like `--hydra-config` (or something like this)
for example, during parser creation you could check `if IS_HYDRA_CONFIG_AVAILABLE` (more examples here - https://github.com/catalyst-team/catalyst/blob/master/catalyst/tools/settings.py) and add extra params and their parsing logic. this solution looks good for me, because it provides extra features without breaking changes

Catalyst Config API **must** save it backward compatibility - it's our production-ready solution :)

Scitator: @omry could you please provide more examples?
for example, I am not sure, that ``cfg.foo.bar instead of cfg["foo"]["bar"]`` is better solution - I personally prefer the second variant because it shows, that you are working with the config

other features looks great, but it examples are much better :)

Scitator (dismissed):

omry:
@Scitator,
I want to preface that I think if you are not convinced that using OmegaConf/Hydra with Catalyst is a good thing please do not accept this PR.
Benefitting from what Hydra has to offer will take a strategic decision and not a feature creep by user provided pull requests.

Having said that:
OmegaConf is well documented and has plenty of examples.
I recommend that you go over the [docs](https://omegaconf.readthedocs.io/) or the [slides](https://docs.google.com/presentation/d/e/2PACX-1vT_UIV7hCnquIbLUm4NnkUpXvPEh33IKiUEvPRF850WKA8opOlZOszjKdZ3tPmf8u7hGNP6HpqS-NT5/pub?start=false&loop=false&delayms=3000&slide=id.p).

> Catalyst Config API must save it backward compatibility - it's our production-ready solution :)
Here is a thought:
What is internally Catalyst worked with OmegaConf but still allowed users to provide configs as dictionaries?
You can convert a dictionary to OmegaConf very easily:
```
cfg = OmegaConf.create({"foo": "bar"}).
assert cfg["foo"] == "bar"
assert cfg.foo == "bar"
```
You asked about types:
See Structured Configs in the docs. Typing support in OmegaConf is very advanced.

Hydra adds more powerful command line that support type casting in the command line ([basic](https://hydra.cc/docs/advanced/override_grammar/basic), [extended](https://hydra.cc/docs/advanced/override_grammar/extended)).

Scitator (changes requested):
sorry, but you need to merge with the master branch correctly - too many deletions

catalyst/dl/experiment/hydra_config.py
```diff
@@ -0,0 +1,553 @@
...
+        return criterion
+
+    def _get_optimizer(
+        self,
+        stage: str,
+        models: Dict[str, Model],
+        name: str,
+        params: DictConfig,
+    ) -> Optimizer:
+        # getting layer-wise parameters
```
Scitator, line 133: @bagxi looks like we need to move this to the utils, what do you think?

catalyst/dl/runner/hydra_supervised.py
```diff
@@ -0,0 +1,202 @@
...
+import torch
+
+from catalyst.dl.experiment.hydra_config import HydraConfigExperiment
+from catalyst.dl.runner.runner import Runner
+from catalyst.tools.typing import Device, RunnerModel
+
+logger = logging.getLogger(__name__)
+
+
+class HydraSupervisedRunner(Runner):
```
Scitator, line 14: why do we need it?

Scitator (changes requested):
could you please rebase your changes? it's hard to understand, what is going here with too many extra additions/deletions

PR #949, opened 2020-10-03
Fix unfair split in mnist metric learning dataset

julia-shenshina:
## Before submitting

- [x] Did you update the [CHANGELOG](https://github.com/catalyst-team/catalyst/CHANGELOG.md)?

## Type of Change

## PR review

### DIFF
--- a/catalyst/contrib/datasets/mnist.py
+++ b/catalyst/contrib/datasets/mnist.py
@@ -157,24 +157,8 @@ class MNIST(Dataset):
             os.path.join(self.processed_folder, self.test_file)
         )
 
-    def download(self):
-        """Download the MNIST data if it doesn't exist in processed_folder."""
-        if self._check_exists():
-            return
-
-        os.makedirs(self.raw_folder, exist_ok=True)
-        os.makedirs(self.processed_folder, exist_ok=True)
-
-        # download files
-        for url, md5 in self.resources:
-            filename = url.rpartition("/")[2]
-            download_and_extract_archive(
-                url, download_root=self.raw_folder, filename=filename, md5=md5
-            )
-
-        # process and save as torch files
-        print("Processing...")
-
+    def _train_test_split(self) -> None:
+        """Split the whole dataset to train and test sets."""
         training_set = (
             read_image_file(
                 os.path.join(self.raw_folder, "train-images-idx3-ubyte")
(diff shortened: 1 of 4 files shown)


catalyst/contrib/datasets/mnist.py
```diff
@@ -209,19 +209,74 @@ def extra_repr(self):
...
+    ]
+
+    def __init__(self, root: str, train: bool = True, **kwargs):
+        """
+
+        Args:
+            root: root directory of the dataset
+            train: for MnistMLDataset should always be True
+
+        Raises:
```
AlekseySh, line 231: I think we should remove `train` from args. and consider 2 situations:
`train` is in `kwargs`, then we should check that train is True, otherwise raise an Error 
`train` is not in `kwargs`, then we should add it as `train=True`

in current implementation we allow to pass `train` value directly to init, but don't allow to change what, this is confusing behaviour

catalyst/contrib/datasets/mnist.py
```diff
@@ -277,5 +345,15 @@ def query_size(self) -> int:
         """Query Gallery dataset should have query_size property"""
         return self._query_size
 
+    @property
+    def data(self) -> torch.Tensor:
```
AlekseySh, line 343: why do we need it?

catalyst/contrib/datasets/mnist.py
```diff
@@ -277,5 +345,15 @@ def query_size(self) -> int:
...
+        """Images from MNIST"""
+        return self._mnist.data
+
+    @property
+    def targets(self) -> torch.Tensor:
```
AlekseySh, line 348: why do we need it?

AlekseySh (dismissed):
looks good

catalyst/contrib/datasets/mnist.py
julia-shenshina: why not?
i think it's ok when you can get data and target from your dataset directly without using private fields
moreover, it gives us some coherence in mnist datasets (train mnist dataset has these properties, classic mnist has too)

catalyst/contrib/datasets/mnist.py
julia-shenshina: ~~what is you concern about? 
something like this?
`kwargs = {"train": False}
dataset = MnistMLDataset("./data", True, **kwargs)`
it will fail with exception about multiple values for train args.~~

I thougt it to be a good idea to remind in docs that train should always be True in this case 🤔

julia-shenshina: oh, i think i understood, ok

tests/_tests_scripts/dl_z_mvp_mnist_metric_learning.py
```diff
@@ -29,7 +29,7 @@ def run_ml_pipeline(sampler_inbatch: data.IInbatchTripletSampler) -> float:
         root=dataset_root, train=True, download=True, transform=transforms,
     )
     sampler = data.BalanceBatchSampler(
-        labels=dataset_train.get_labels(), p=10, k=10
+        labels=dataset_train.get_labels(), p=5, k=10
```
Scitator, line 32: btw, could you please also change this example into `Readme.md`?

Scitator (dismissed):
we need to update readme also :)

Scitator (approved):

PR #1176, opened 2021-04-14
Add support for WandbLogger

AyushExel:
## Before submitting (checklist)

- [x] Did you check that your code passes the unit tests `pytest .` ? 
- [x] Did you update the [CHANGELOG](https://github.com/catalyst-team/catalyst/blob/master/CHANGELOG.md)?
- [ ] Did you run [colab minimal CI/CD](https://colab.research.google.com/drive/1JCGTVvWlrIsLXMPRRRSWiAstSLic4nbA) with `latest` and `minimal` requirements?

## Description
This PR adds support for WandbLogger that enables logging metrics and media to W&B dashboard

## Type of Change

## PR review

## Additional Deatils:
* The minimum tests colab seemed stuck while running tests, but the tests passed on my machine. I'll update this thread with test results
* I've made this draft PR as I still want to confirm the hyperparameter logging behavior. On running the `test_finetune2.py` train method, no hyperparameters are being logged. 
### Test logs
Code style
--> `catalyst-make-codestyle && catalyst-check-codestyle`
```
python3.8/site-packages/isort/settings.py:619: UserWarning: Failed to pull configuration information from /home/saksham/Desktop/catalyst/setup.cfg
  warn(f"Failed to pull configuration information from {potential_config_file}")
Skipped 55 files
python3.8/site-packages/isort/main.py:1000: UserWarning: W0501: The following deprecated CLI flags were used and ignored: --apply!
  warn(python3.8/site-packages/isort/main.py:1004: UserWarning: W0500: Please see the 5.0.0 Upgrade guide: https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0/
  warn(
All done! ✨ 🍰 ✨
350 files left unchanged.
python3.8/site-packages/isort/settings.py:619: UserWarning: Failed to pull configuration information from /home/catalyst/setup.cfg
  warn(f"Failed to pull configuration information from {potential_config_file}")
Skipped 55 files
All done! ✨ 🍰 ✨
350 files would be left unchanged.
Failed to pull configuration information from home/catalyst/setup.cfg
0
```
Docs check
-->`rm -rf ./builds; REMOVE_BUILDS=0 make check-docs`
```
reading sources... [100%] tutorials/ddp                                                                                                                                                              
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] tutorials/ddp                                                                                                                                                               
generating indices...  genindex py-modindexdone
highlighting module code... [100%] torch.utils.data.sampler                                                                                                                                          
writing additional pages...  search/home/saksham/anaconda3/envs/catalyst_dev/lib/python3.8/site-packages/catalyst_sphinx_theme/search.html:21: RemovedInSphinx30Warning: To modify script_files in the theme is deprecated. Please insert a <script> tag directly in your theme instead.
  {% trans %}Please activate JavaScript to enable the search
done
copying static files... ... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in builds.
#### CODE: 0 ####
```
Tests
--> `pytest .`
```
337 passed, 134 skipped, 2 xfailed, 93 warnings in 439.09s (0:07:19)
```
@Scitator Let me know if I missed  any steps here 
### FAQ
Please review the FAQ before submitting an issue:
- [ ] I have read the [documentation and FAQ](https://catalyst-team.github.io/catalyst/)
- [ ] I have reviewed the [minimal examples section](https://github.com/catalyst-team/catalyst#minimal-examples)
- [ ] I have checked the [changelog](https://github.com/catalyst-team/catalyst/blob/master/CHANGELOG.md) for main framework updates
- [ ] I have read the [contribution guide](https://github.com/catalyst-team/catalyst/blob/master/CONTRIBUTING.md)
- [ ] I have joined [Catalyst slack (#__questions channel)](https://join.slack.com/t/catalyst-team-core/shared_invite/zt-d9miirnn-z86oKDzFMKlMG4fgFdZafw) for issue discussion


catalyst/loggers/wandb.py
```diff
@@ -0,0 +1,180 @@
...
+    def log_hparams(
+        self,
+        hparams: Dict,
+        scope: str = None,
+        # experiment info
+        run_key: str = None,
+        stage_key: str = None,
+    ) -> None:
+        """Logs hyperparameters to the logger."""
+        self.run.config.update(hparams)
```
Scitator, line 169: would that update the config on the Wandb side?

catalyst/loggers/wandb.py
```diff
@@ -0,0 +1,180 @@
...
+                metrics=metrics, step=global_batch_step, loader_key=loader_key, suffix="/batch"
+            )
+        elif scope == "epoch":
+            for loader_key, per_loader_metrics in metrics.items():
+                self._log_metrics(
+                    metrics=per_loader_metrics,
+                    step=global_epoch_step,
+                    loader_key=loader_key,
+                    suffix="/epoch",
+                )
```
Scitator: we have an update for tensorboard logger recently (we log loader metric just after the loader rather then in the end of the epoch): https://github.com/catalyst-team/catalyst/blob/master/catalyst/loggers/tensorboard.py#L122#L136
would you like to do the same for wandb one?

catalyst/loggers/wandb.py
```diff
@@ -0,0 +1,180 @@
...
+        loader_batch_len: int = 0,
+        loader_sample_len: int = 0,
+        loader_batch_step: int = 0,
+        loader_sample_step: int = 0,
+    ) -> None:
+        """Logs batch and epoch metrics to wandb."""
+        if scope == "batch":
+            metrics = {k: float(v) for k, v in metrics.items()}
+            self._log_metrics(
+                metrics=metrics, step=global_batch_step, loader_key=loader_key, suffix="/batch"
```
Scitator:
```suggestion
                metrics=metrics, step=global_sample_step, loader_key=loader_key, suffix="/batch"
```
for convenient logging :)

catalyst/tests/test_finetune2.py
```diff
@@ -30,6 +30,7 @@ def get_loggers(self):
             "csv": dl.CSVLogger(logdir=self._logdir),
             "tensorboard": dl.TensorboardLogger(logdir=self._logdir),
             "mlflow": dl.MLflowLogger(experiment=self._name),
+            "wandb": dl.WandbLogger(project="catalyst_test", config=self.hparams),
```
Scitator: could you please wrap it with `if-else` like `mlflow` one in master branch?

Scitator (commented):
The logger looks great! After a few noted hotfixes, let's update the branch to the current master and CI test and merge it 👍

catalyst/loggers/wandb.py
AyushExel: Yes. But this is not recommended. I'll comment it out. config when updated, should ideally only add new keys and should avoid updating the old keys which isn't guaranteed here. That's what I wanted to discuss regarding config. The recommended way here would be to pass config hparams in the construntor as:
`logger['wandb'] = WandbLogger(project='name, config=self.harams)`

But anyway neither of these are updating any values in config as it stays empty when running test_finetune2.py

catalyst/tests/test_finetune2.py
AyushExel: Yup done

catalyst/loggers/wandb.py
AyushExel: Okay, I tried this but as I mentioned on slack, wandb needs steps to be strictly increasing, we cannot go back to previous steps. And there's only one global step. So, doing it this way will cause it to drop values, so I'm  using `global_epoch_step` for all cases as it ensures that it's strictly increasing

AyushExel:
@Scitator I'm not sure why tests are failing. I don't think the changes should cause it. Here's the log of pytest on my machine from this branch:
```
336 passed, 132 skipped, 2 xfailed, 92 warnings in 450.74s (0:07:30) 
```

ditwoo (dismissed):

catalyst/tests/test_finetune2.py
```diff
@@ -31,7 +31,9 @@ def get_loggers(self):
             "tensorboard": dl.TensorboardLogger(logdir=self._logdir),
         }
         if SETTINGS.ml_required:
```
Scitator:
```suggestion
        if SETTINGS.mlflow_required:
```
looks like here is a typo

catalyst/loggers/wandb.py
Scitator: 🤔 could we move the initial init here? from the framework point of view, it's the only correct place to store the hyperparameters – during `logger.log_hparams` rather that logger initialization.

catalyst/loggers/wandb.py
Scitator: all `*_step` counters are always increasing during the Catalyst run, so that should be okay ;)

catalyst/loggers/wandb.py
Scitator: The WandbLogger looks great!
Let's resolve this issue and merge it to the master branch.
I think, we could return `self.run.config.update(hparams)` to `log_hparams` and update the docs with the ideal use case scenario: pass the experiment hyperparameter through `log_hparams` rather than `init`.

catalyst/loggers/wandb.py
AyushExel: Yeah, the individual `*_step` counters are increasing but wandb run only has one global counter. Consider a scenario where `global_step` is 3 and `sample_step` is 2. I tested with this and the values were being dropped with warnings.

catalyst/loggers/wandb.py
AyushExel: Done :slightly_smiling_face:

AyushExel:
@Scitator something that I just noticed. As discussed earlier, I'm sticking with using `global_epoch_step` for each metric as wandb supports only one global step for all metrics and that should be strictly increasing. But even in the current settings I'm getting these warnings. ( Testing on `test/test_finetune2.py's  train_experiment("cuda:0")`)
Screenshot from 2021-04-26 13-10-57
Now, I remember that this worked fine the last time when I marked this PR for review( [example output](https://wandb.ai/cayush/catalyst_test/runs/28he4eoj?workspace=user-cayush)). It is strange as only the global epoch step is being used and it should definitely be increasing unless there's a typo. Can you find something out of place? Or maybe a recent commit introduced this?

Scitator:
@AyushExel could you please try use `_suffix` rather than `/suffix` for the logging? maybe different names for the batch-based and epoch-based metrics would work for Wandb

AyushExel:
@Scitator sorry it was a false alarm. I didn't rebuild the package after changing `sample_step` to `global_step`. The logger works fine and I've updated the test to not log hyperparameters when initializing, instead it only logs through `log_hparams` now. Should be good to go
EDIT : Example [run](https://wandb.ai/cayush/catalyst_test/runs/1fgcecfd?workspace=user-cayush)

catalyst/loggers/wandb.py
```diff
@@ -0,0 +1,180 @@
...
+        loader_batch_len: int = 0,
+        loader_sample_len: int = 0,
+        loader_batch_step: int = 0,
+        loader_sample_step: int = 0,
+    ) -> None:
+        """Logs batch and epoch metrics to wandb."""
+        if scope == "batch":
+            metrics = {k: float(v) for k, v in metrics.items()}
+            self._log_metrics(
+                metrics=metrics, step=global_epoch_step, loader_key=loader_key, suffix="/batch"
```
Scitator:
```suggestion
                metrics=metrics, step=global_sample_step, loader_key=loader_key, suffix="/batch"
```

Scitator: as far as this is per-batch/sample statistics we have to use per-sample counter

AyushExel: Using `global_sample_step` will cause this problem -> https://github.com/catalyst-team/catalyst/pull/1176#issuecomment-826594360

Scitator: @AyushExel could you please try using _suffix rather than /suffix for the logging? maybe different names for the batch-based and epoch-based metrics would work for Wandb

you could not write per-**_batch_** metrics under the **_epoch_** counter - it's simply incorrect behaviour
could you please try using different names for per-batch and per-epoch metrics?

AyushExel: Yeah, the step counter behavior is pretty strict which comes up a lot. But I'm pretty sure that's how it is right now. I tried changing names as you suggested but that didn't work. See [this ](https://docs.wandb.ai/guides/track/log#incremental-logging)where this behaviour is explained. 

> As long as you keep passing the same value for step, W&B will collect the keys and values from each call in one unified dictionary. As soon you call wandb.log() with a different value for step than the previous one, W&B will write all the collected keys and values to the history, and start collection over again. Note that this means you should only use this with consecutive values for step: 0, 1, 2, .... This feature doesn't let you write to absolutely any history step that you'd like, only the "current" one and the "next" one.

ditwoo:
Hi, @AyushExel 

Could you please check that metrics, parameters, etc. will be logged as expected when the experiment has multiple stages?

AyushExel:
> Hi, @AyushExel
> 
> Could you please check that metrics, parameters, etc. will be logged as expected when the experiment has multiple stages?

@Ditwoo any specific example in tests that I should run the logger with ?

ditwoo:
> > Hi, @AyushExel
> > Could you please check that metrics, parameters, etc. will be logged as expected when the experiment has multiple stages?
> 
> @Ditwoo any specific example in tests that I should run the logger with ?

I think you can adjust an example from readme:

```python
import os
from torch import nn, optim
from torch.utils.data import DataLoader
from catalyst import dl, utils
from catalyst.contrib.datasets import MNIST
from catalyst.data.transforms import ToTensor


class CustomRunner(dl.IRunner):
    def __init__(self, logdir, device):
        super().__init__()
        self._logdir = logdir
        self._device = device

    def get_engine(self):
        return dl.DeviceEngine(self._device)

    def get_loggers(self):
        return {
            "console": dl.ConsoleLogger(),
            "csv": dl.CSVLogger(logdir=self._logdir),
            # TODO: finish this part
            "wandb": dl.WandbLogger(...),
        }

    @property
    def stages(self):
        return ["train_freezed", "train_unfreezed"]

    def get_stage_len(self, stage: str) -> int:
        return 3

    def get_loaders(self, stage: str):
        loaders = {
            "train": DataLoader(
                MNIST(os.getcwd(), train=True, download=True, transform=ToTensor()), batch_size=32
            ),
            "valid": DataLoader(
                MNIST(os.getcwd(), train=False, download=True, transform=ToTensor()), batch_size=32
            ),
        }
        return loaders

    def get_model(self, stage: str):
        model = (
            self.model
            if self.model is not None
            else nn.Sequential(nn.Flatten(), nn.Linear(784, 128), nn.ReLU(), nn.Linear(128, 10))
        )
        if stage == "train_freezed":
            # freeze layer
            utils.set_requires_grad(model[1], False)
        else:
            utils.set_requires_grad(model, True)
        return model

    def get_criterion(self, stage: str):
        return nn.CrossEntropyLoss()

    def get_optimizer(self, stage: str, model):
        if stage == "train_freezed":
            return optim.Adam(model.parameters(), lr=1e-3)
        else:
            return optim.SGD(model.parameters(), lr=1e-1)

    def get_scheduler(self, stage: str, optimizer):
        return None

    def get_callbacks(self, stage: str):
        return {
            "criterion": dl.CriterionCallback(
                metric_key="loss", input_key="logits", target_key="targets"
            ),
            "optimizer": dl.OptimizerCallback(metric_key="loss"),
            "checkpoint": dl.CheckpointCallback(
                self._logdir, loader_key="valid", metric_key="loss", minimize=True, save_n_best=3
            ),
        }

    def handle_batch(self, batch):
        x, y = batch
        logits = self.model(x)

        self.batch = {
            "features": x,
            "targets": y,
            "logits": logits,
        }

runner = CustomRunner("./logs", "cpu")
runner.run()
```

Or adjust [this test file](https://github.com/catalyst-team/catalyst/blob/master/catalyst/tests/test_finetune.py).

catalyst/loggers/wandb.py
Scitator: Then I think you could use different names for batch-based and epoch-based metrics – just rename them in the way, that Wandb would understand :)

AyushExel: Yes this might make more sense. So, instead of suffix, I think we can use the prefix. If you log anything in wandb as `{'prefix/key': value}`, it creates it's own section in the dashboard called `prefix` and all metrics belonging to the same prefix share the same section.

AyushExel:
@Ditwoo Yes above mentioned examples work. Here's sample [dashboard](https://wandb.ai/cayush/catalyst_integration?workspace=user-cayush). something that I've been noticing in these examples is that they're passing empty Dicts in `log_hparams` function call. Is that an intended use case or something that's still in the works?

catalyst/loggers/wandb.py
Scitator: it's up to you, you could create `{metric_name}_{prefix}/{somethin}`, so it would be accuracy_epoch and accuracy_batch

ditwoo:
> `log_hparams`

As far as I know, `hparams` used when you are doing some architecture search (using optuna or similar).
@Scitator could you please comment if it's expected behavior for `hparams`?

Scitator:
It's better to refer to the docs: https://catalyst-team.github.io/catalyst/api/core.html?highlight=hparams#catalyst.core.runner.IRunner.hparams.

catalyst/loggers/wandb.py
AyushExel: Okay updated the names of the metrics as you suggested. Now each metric has its section like `{metric_name}_{prefix}`. I think it's all good now.

AyushExel:
Okay I think then it's all good. `Runner.hparams` for these experiments is empty so it makes sense that no values are logged.

Scitator:
@AyushExel could you please update the changelog (link in the PR description)?

Scitator:
btw, you also need to add the Logger to the [docs](https://github.com/catalyst-team/catalyst/blob/master/docs/api/loggers.rst)

Scitator:
final changes: could you please merge with master?

PR #1220, opened 2021-05-25
Adding ProfilerCallback

eloygeenjaar:
## Before submitting (checklist)

- [ ] Did you check that your code passes the unit tests `pytest .` ? 
- [x] Did you update the [CHANGELOG](https://github.com/catalyst-team/catalyst/blob/master/CHANGELOG.md)?
- [ ] Did you run [colab minimal CI/CD](https://colab.research.google.com/drive/1JCGTVvWlrIsLXMPRRRSWiAstSLic4nbA) with `latest` and `minimal` requirements?

## Type of Change

## PR review

### FAQ
Please review the FAQ before submitting an issue:
- [x] I have read the [documentation and FAQ](https://catalyst-team.github.io/catalyst/)
- [x] I have reviewed the [minimal examples section](https://github.com/catalyst-team/catalyst#minimal-examples)
- [x] I have checked the [changelog](https://github.com/catalyst-team/catalyst/blob/master/CHANGELOG.md) for main framework updates
- [x] I have read the [contribution guide](https://github.com/catalyst-team/catalyst/blob/master/CONTRIBUTING.md)
- [x] I have joined [Catalyst slack (#__questions channel)](https://join.slack.com/t/catalyst-team-core/shared_invite/zt-d9miirnn-z86oKDzFMKlMG4fgFdZafw) for issue discussion


catalyst/callbacks/profiler.py
```diff
@@ -0,0 +1,85 @@
...
+                    logdir="./logs",
+                    valid_loader="valid",
+                    valid_metric="loss",
+                    minimize_valid_metric=True,
+                    verbose=True,
+                    callbacks=[ProfilerCallback(profiler)])
+                print(profiler.key_averages().table(sort_by="cpu_time_total", row_limit=10))
+                print(profiler.key_averages().table(sort_by="self_cpu_memory_usage", row_limit=10))
+                if torch.cuda.is_available():
+                    print(profiler.key_averages().table(sort_by="cuda_time_total", row_limit=10))
```
Scitator, line 63: could you please run `make-catalyst-codestyle` on this code?

catalyst/callbacks/profiler.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+
+    def __init__(self, profiler=None):
```
Scitator, line 66: let's also add an option to create `profiler` by params

Scitator: I mean, it would be also great to get ProfilerCallback example with Config API... something like
Config API - minimal example in https://github.com/catalyst-team/catalyst#minimal-examples

Scitator (commented):
👍

catalyst/callbacks/profiler.py
eloygeenjaar: If we create the `profiler` using params, we won't be able to wrap the training code in a `with profiler:` statement, or we would have to return the initialized profiler, but that seems over the top since the user could just instantiate the profiler by themselves in that case (which is the current implementation)
.

Scitator: you could use `self.profiler.__enter__()` to emulate with statement, something like:
```python
class ProfilerCallback:
  init(self, profiler=None, mode=loader or epoch): 
    profiler = profiler or init
    mode = mode

  on_epoch_start: profiler.__enter__() if mode == epoch else nothing
  on_loader_start: profiler.__enter__() if mode == loader else nothing
  on_loader_end: profiler.__exit__() if mode == loader else nothing
  on_epoch_end: profiler.__exit__() if mode == epoch else nothing
  ```

**Comment by ditwoo** [comment]
Hi, @eloygeenjaar could you please add some code for checking the PyTorch version.
I mean profile callback can be used only with PyTorch >= 1.8.1 version and it would be nice to have this callback with the correct versions of PyTorch.

File: catalyst/callbacks/profiler.py (hunk 2 above)
Comment (eloygeenjaar) [reply catalyst/callbacks/profiler.py:66]: I tried something similar after our discussion in the Catalyst slack, the profiler should be before the dataset instantiation which is at the `on_stage_start()` level

**Comment by eloygeenjaar** [comment]
> Hi, @eloygeenjaar could you please add some code for checking the PyTorch version.
> I mean profile callback can be used only with PyTorch >= 1.8.1 version and it would be nice to have this callback with the correct versions of PyTorch.

Would you prefer I add an assert, raise an exception, or raise a warning that notifies the user that the profiler can not be used for the run. I am assuming you want me to add it in the `profiler.py` file right? Not in the minimal example

File: catalyst/callbacks/profiler.py (hunk 2 above)
Comment (Scitator) [reply catalyst/callbacks/profiler.py:66]: you could add mode=stage, is there any problem with that?

**Comment by ditwoo** [comment]
> > Hi, @eloygeenjaar could you please add some code for checking the PyTorch version.
> > I mean profile callback can be used only with PyTorch >= 1.8.1 version and it would be nice to have this callback with the correct versions of PyTorch.
> 
> Would you prefer I add an assert, raise an exception, or raise a warning that notifies the user that the profiler can not be used for the run. I am assuming you want me to add it in the `profiler.py` file right? Not in the minimal example

I think this should be handled inside `catalyst/catalyst/callbacks/__init__.py`.
I mean you need to import profiler callback only when PyTorch >= 1.8.1, so I think it should be something like this

```python
# somewhere inside 'catalyst/catalyst/callbacks/__init__.py'
if torch.__version__ >= '1.8.1':
    from catalyst.callbacks.profiler import ProfilerCallback
```

catalyst/callbacks/profiler.py
eloygeenjaar: I think the problem was that the dataset instantiation happens before the callbacks are executed, so memory profiling would issue lots of warnings

Scitator:
should be already merged with https://github.com/catalyst-team/catalyst/pull/1226

